PR #13, opened 2024-05-17
9 add postgresql support to manager database access

d-linko:


src/mgr/nm_mgr_sql.c
```diff
@@ -65,13 +66,23 @@ typedef enum db_con_t {
 } db_con_t;
 
 /* Global connection to the MYSQL Server. */
+#ifdef HAVE_MYSQL
```
BrianSipos, line 69: For longer term support, this may be better to split into two files for each SQL type and conditionally include in target library. Would not be a change of code just file location. Thoughts?

BrianSipos (changes requested):

src/mgr/nm_mgr_sql.c
d-linko: yeah that would make sense and would make it easier for adding new database types in the future

BrianSipos (approved):

PR #20, opened 2024-07-23
Fixing tv create issue

d-linko:
fixing timevalue by using `adm_get_parm_tv.secs` for relative times

### DIFF
--- a/src/indep_adms/agent/adm_amp_agent_impl.c
+++ b/src/indep_adms/agent/adm_amp_agent_impl.c
@@ -24,6 +24,7 @@
 #include "../shared/msg/msg.h"
 #include "rda.h"
 #include "ldc.h"
+#include "time.h"
 
 /*   STOP CUSTOM INCLUDES HERE  */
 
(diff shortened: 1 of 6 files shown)


src/indep_adms/agent/adm_amp_agent_impl.c
```diff
@@ -1806,11 +1807,12 @@ tnv_t *amp_agent_ctrl_add_tbr(eid_t *def_mgr, tnvc_t *parms, int8_t *status)
 	rule_t *tbr = NULL;
 
 	ari_t *id = adm_get_parm_obj(parms, 0, AMP_TYPE_ARI);
-	OS_time_t start = amp_tv_to_ctime(adm_get_parm_tv(parms, 1, &success), NULL);
```
BrianSipos, line 1809: Should the start time actually be an absolute time based on a non-null second argument?

BrianSipos: The actual rule handling is conditioned on the start time being relative or absolute, so no special handling is needed in the CTRL conversion here.

BrianSipos (approved):

PR #23, opened 2024-08-15
updated num_parms for ops and check for max eval=0

d-linko:
There is an issue where the internal nm ARI builder was setting all the operator has_parm to true even when they dont actually have parameters. ACE was not setting the has_parm flag to true so sending commands with operatings to the nm from anms was not working. Current fix is by setting the num_parm to 0 when creating an operator. Also fixed a check with max_evals for SBR so now 0 means theres no max

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -14,6 +14,7 @@ on:
 
 jobs:
   unit-test:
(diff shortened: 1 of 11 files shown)


BrianSipos (commented):
The changes seem okay but CI is failing because the testenv sh scripts are using `docker-compose` instead of `docker compose` (sub command with space). Can you update the test scripts?

BrianSipos (approved):

Issue #26, opened 2024-09-05
ARI handling of integer types in text form


BrianSipos:
The CACE library needs to handle untyped and typed integer values in text form according to [Section 4.2.1 of the ARI draft](https://www.ietf.org/archive/id/draft-ietf-dtn-ari-02.html#section-4.2.1).

Tests must include positive verification of encoding and decoding specific integer values. This includes alternatives for ARI types BYTE, INT, UINT, VAST, and UVAST and alternatives for encoding using base-2, base-10, base-16 and alternatives for leading signs (including encoded values like "+1" for unsigned types).

justinethier:
Closing, this has been dealt with already by multiple PR's.

PR #28, opened 2024-09-09
Issue #26 - Decode binary number ARI with optsign

justinethier:
Handle decoding of binary number ARI's that include the `+` optsign.

Also includes additional related test cases for completeness.


BrianSipos (approved):
Changes look good. I don't know why CI is failing here.

BrianSipos:
I rebased and fixed the CI so squash when ready.

PR #46, opened 2024-09-16
ARI handling updates for floating point types in text form

justinethier:
Support full set of hexadecimal floats per the ARI document.

Added additional test cases for typed floating point types, and range tests for `REAL32`.


src/cace/ari/text_val.lex
```diff
@@ -130,7 +130,7 @@ DECDIG [0-9]
     }
 }
 
-<PRIMITIVE,LT_ANYFLOAT>[+-]?(({DECDIG}+|{DECDIG}*\.{DECDIG}*)([eE][+-]?{DECDIG}+)|0[xX]{HEXDIG}\.{HEXDIG}*([pP][+-]?{DECDIG})|{DECDIG}*(\.{DECDIG}*)?|Infinity)|NaN {
+<PRIMITIVE,LT_ANYFLOAT>[+-]?(({DECDIG}+|{DECDIG}*\.{DECDIG}*)([eE][+-]?{DECDIG}+)|0[xX]({HEXDIG}+|{HEXDIG}*\.{HEXDIG}*)([pP][+-]?{DECDIG})|0[xX]{HEXDIG}*(\.{HEXDIG}*)?|{DECDIG}*(\.{DECDIG}*)?|(?i:Infinity))|(?i:NaN) {
```
BrianSipos, line 133: This hex pattern seems overly complex; does it agree with the ABNF that is in the ARI spec?

justinethier: Right, the ABNF from the ARI spec is:

```
float-hex = optsign "0x" 1*HEXDIG "." *HEXDIG
            ["p" optsign 1*DIGIT]
```

The pattern from the lexer is broken into two parts to allow for the exponent to be optional. This was inspired by the decimal matcher which is written in the same way:

```
0[xX]({HEXDIG}+|{HEXDIG}*\.{HEXDIG}*)([pP][+-]?{DECDIG})
0[xX]{HEXDIG}*(\.{HEXDIG}*)?
```

This form matches the ARI spec and goes one step further allowing the decimal point to be optional (there is a ticket on the ARI spec's repo discussing this further).

We might be able to simplify the hex logic into a single rule. I ran into problems when doing that initially, and fell back to using two rules like the decimal matcher.

justinethier: @BrianSipos - The hex pattern has been simplified to the following, while still retaining all of the functionality noted above:

```
0[xX]({HEXDIG}+|{HEXDIG}*\.{HEXDIG}*)([pP][+-]?{DECDIG})?
```

BrianSipos (approved):

PR #47, opened 2024-09-25
Test cases to support ARI handling of string types in text form

justinethier:
Adding additional test cases to support TEXTSTR and BYTESTR types.

Also addressed FIXME comments that required percent encoding of escape characters.


BrianSipos (approved):
Good catches and test case fixes, this also caught a memory leak which is also good.

PR #48, opened 2024-09-27
ARI handling updates for containers in text form

justinethier:
This MR adds addresses ARI handling of AC, AM, and TBL objects:

- Updated parsing of literal parameter lists to support case-insensitive keys and to raise an error if a duplicate key is found. This change is applicable to TBL objects as well as other objects that use parameter lists. 
- Added unit tests for the above.
- Added dedicated decoding test cases for all three object types.
- Added additional unit tests to handle edge cases with AC and TBL objects.

In addition, this PR adds the ability to handle object type names in a case insensitive manner and provides tests for them.


src/cace/ari/text_util.h
```diff
@@ -140,6 +140,22 @@ int slash_unescape(string_t out, const cace_data_t *in);
  */
 void strip_space(string_t out, const char *in, size_t in_len);
 
+/** Convert a text string to lowercase.
+ * This is written to work on byte strings, not unicode.
+ *
+ * @param[out] out The output buffer, which will be replaced.
+ * @return Zero upon success.
+ */
+int cace_string_tolower(string_t out);
```
BrianSipos, line 149: If it really cannot fail then these can return void and use precond macro `CHKVOID()`. Since the return values were not being checked anyway.

src/cace/ari/text_util.h
```diff
@@ -140,6 +140,22 @@ int slash_unescape(string_t out, const cace_data_t *in);
...
+
+/** Convert a text string to uppercase.
+ * This is written to work on byte strings, not unicode.
+ *
+ * @param[out] out The output buffer, which will be replaced.
+ * @return Zero upon success.
+ */
+int cace_string_toupper(string_t out);
```
BrianSipos, line 157: Same as other function.

BrianSipos (changes requested):
Other than the new return types, this looks good

src/cace/ari/text_util.h
justinethier: Thanks @BrianSipos, changed both functions to `void`.

src/cace/ari/text_util.h
justinethier: Updated this one as well, thanks.

BrianSipos (approved):

PR #49, opened 2024-10-03
ARI handling updates for CBOR in text form

justinethier:
Remove CBOR extended diagnostic notation as it is not practical to add support for parsing EDN.

Add additional unit tests for ARI's containing typed CBOR.

### DIFF
--- a/test/cace/test_ari_text.c
+++ b/test/cace/test_ari_text.c
@@ -19,6 +19,7 @@
  * Test the ari_text.h interfaces.
  */
 #include <cace/ari/text.h>
+#include <cace/ari/text_util.h>
 #include <cace/amm/typing.h>
 #include <unity.h>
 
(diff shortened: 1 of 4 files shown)


test/cace/test_ari_text.c
```diff
@@ -606,6 +606,35 @@ void test_ari_text_decode_lit_prim_bstr(const char *text, const char *expect, si
     ari_deinit(&ari);
 }
 
+TEST_CASE("ari:/CBOR/h''", NULL, 0)
+TEST_CASE("ari:/CBOR/h'A164746573748203F94480'", "\xA1\x64\x74\x65\x73\x74\x82\x03\xF9\x44\x80", 11)
+TEST_CASE("ari:/CBOR/h'A1%2064%2074%2065%2073%2074%2082%2003%20F9%2044%20%2080'", "\xA1\x64\x74\x65\x73\x74\x82\x03\xF9\x44\x80", 11)
+void test_ari_text_decode_lit_typed_cbor(const char *text, const char *expect, size_t expect_len)
```
BrianSipos, line 612: Minor suggestion to have expect be in hex and decode to byte string similar to the byte string handling. For your vectors this is not a problem but some could have embedded zero bytes.

BrianSipos (commented):
One suggested fixture change but otherwise good to merge.

test/cace/test_ari_text.c
justinethier: Good catch @BrianSipos - updated to allow vectors that contain embedded `0x00` bytes.

justinethier: @BrianSipos Are you good with the latest changes for this PR?

BrianSipos: I had originally meant to use a direct hex-encoded binary as the expected value but didn't explain exactly. I just made a commit to adjust that and simplify the expected parameters. I recommend to follow a similar pattern in other places if your expected values are binary. Not the only way but I think it makes the expected value more human readable.

BrianSipos (approved):

test/cace/test_ari_text.c
justinethier: Good deal. Agree this is more readable and will keep in mind for upcoming CBOR tests. Thanks @BrianSipos!

PR #51, opened 2024-10-08
Proposed changes for adding type checking to RPTSET parameter list values

justinethier:
This PR adds proposed type checking for RPTSET parameter list values.

@BrianSipos please advise if you would prefer to go a different route with this.


BrianSipos:
> This PR adds proposed type checking for RPTSET parameter list values.
> 
> @BrianSipos please advise if you would prefer to go a different route with this.

If possible, it could be better to use `amm_type_convert(amm_type_get_builtin(ARI_TYPE_TP))` and similar to make conversions here rather than requiring one strict form for the inputs. With the status result from the conversion used to determine whether the input is valid.
Do you mind prototyping this? And if it works the commit could be added to this PR.

BrianSipos (commented):
Concept is good, using type conversion instead of exact match may be easier on the user.

justinethier:
@BrianSipos Added prototype commits for TD/TP (call roll back from PR if needed). What do you think?

src/cace/amm/typing.c
```diff
@@ -523,6 +523,74 @@ static int builtin_real64_convert(const amm_type_t *self, ari_t *out, const ari_
...
+        case ARI_PRIM_FLOAT64:
+        {
+            double v = obj->value.as_float64;
+            if (isnan(v) || isinf(v) || v > INT64_MAX || v < 0.0)
+            {
+              return 1;
+            }
+
+            double integral;
+            double frac = modf(v, &integral);
```
BrianSipos, line 549: Can you include some unit tests (in `test_typing.c`) for this behavior with negative values? It might need a little adjusting but test should confirm. I didn't realize that the time-specific conversion was missing earlier so this is a good catch.

src/cace/ari/text_str.yac
```diff
@@ -262,7 +263,18 @@ ssp-literal:
             }
             else
             {
-                //FIXME where to convert (and check) type?
+              ari_t tmp;
+              ari_init(&tmp);
+              int err = amm_type_convert(amm_type_get_builtin(ARI_TYPE_TP), &tmp, found_r);
+              if (err) 
+              {
+                ari_deinit(found_r);
```
BrianSipos, line 271: Please add `ari_deinit(&tmp)` for consistency

src/cace/ari/text_str.yac
```diff
@@ -262,7 +263,18 @@ ssp-literal:
...
+                found_r = NULL;
+              }
+              else
+              {
+                ari_set_copy(found_r, &tmp);
```
BrianSipos, line 276: This can be `ari_set_move()` to avoid needing to deinit the `tmp` var.

src/cace/ari/text_str.yac
```diff
@@ -401,7 +413,18 @@ report:
             }
             else
             {
-                //FIXME where to convert (and check) type?
+                ari_t tmp;
+                ari_init(&tmp);
+                int err = amm_type_convert(amm_type_get_builtin(ARI_TYPE_TD), &tmp, found_t);
+                if (err) 
+                {
+                  ari_deinit(found_t);
```
BrianSipos, line 421: Same two comments as above for this use.

BrianSipos (changes requested):
Very minor feedback in comments. These are very good fixes and updates.

src/cace/ari/text_str.yac
justinethier: No problem, added this call.

src/cace/ari/text_str.yac
justinethier: I changed the call, however I noticed the code in `ari_set_move` is:

```
// reset the state of the src (not deinit)
ari_reset(src);
```

It may not matter in this case, but in general it would seem `ari_deinit` still needs to be called after `ari_set_move`?

src/cace/amm/typing.c
justinethier: Good call! Added test cases and also fixed the return value when invalid floating point numbers (including negatives) are found.

justinethier:
@BrianSipos This is ready when you have a chance to take another look. Do have one outstanding question above regarding `ari_set_move` and `ari_deinit`.

test/cace/test_amm_typing.c
```diff
@@ -288,3 +288,33 @@ void test_amm_type_convert_ident(const char *inhex, const char *expecthex)
     const amm_type_t *type = amm_type_get_builtin(ARI_TYPE_IDENT);
     check_convert(type, inhex, expecthex);
 }
+
+TEST_CASE("820C1A2B450625", "820C1A2B450625") // ari:/TP/20230102T030405Z
+TEST_CASE("82041864", "820C1864") // ari://INT/100 -> ari://TP/100
+TEST_CASE("8208FB40593F34D6A161E5", "820C82231A000F68D4") // ari://REAL64/100.9876
+TEST_CASE("8209FA5F000000", NULL) // ari://REAL64/0x7FFFFFFFFFFFFFFFF --> exceeds INT64 MAX
+TEST_CASE("820D1864", "820C1864") // ari://TD/100 -> ari://TP/100
```
BrianSipos, line 296: This one should actually fail. The same with the other direction to TD. Both can convert from primitive, int, float values but not between each other. I think the negative int case below should succeed.

test/cace/test_amm_typing.c
```diff
@@ -288,3 +288,33 @@ void test_amm_type_convert_ident(const char *inhex, const char *expecthex)
...
+{
+    const amm_type_t *type = amm_type_get_builtin(ARI_TYPE_TP);
+    check_convert(type, inhex, expecthex);
+}
+
+TEST_CASE("820D822018CD", "820D822018CD") // ari:/TD/PT20.5S
+TEST_CASE("82041864", "820D1864") // ari://INT/100 -> ari://TD/100
+TEST_CASE("8208FB40593F34D6A161E5", "820D82231A000F68D4") // ari://REAL64/100.9876
+TEST_CASE("8209FA5F000000", NULL) // ari://REAL64/0x7FFFFFFFFFFFFFFFF --> exceeds INT64 MAX
+TEST_CASE("820C1864", "820D1864") // ari://TP/100 -> ari://TD/100
```
BrianSipos, line 311: Same as above comment. And if this is unclear from the spec, the spec needs more explanation.

src/cace/ari/text_str.yac
BrianSipos: This is a good point the deinit is still needed after a move from. Is the doxygen comment for move funce clear about this? Might need an update if not.

justinethier: Updated doxygen comments and added deinit calls.

src/cace/amm/typing.c
justinethier: That's fair, modified to allow negative values, and adjusted test cases accordingly.

test/cace/test_amm_typing.c
justinethier: Thanks for clarifying, wasn't sure if we wanted to allow TD/TP conversions. Updated accordingly and also fixed the negative int case as well.

test/cace/test_amm_typing.c
justinethier: Address these as well.

justinethier:
@BrianSipos Addressed comments from your latest review. Please let me know if you have more questions/concerns.

BrianSipos:
> @BrianSipos Addressed comments from your latest review. Please let me know if you have more questions/concerns.

This is not a code issue but I notice your text ARIs in C comments have bad syntax. The two slash "ari://" is for object reference values not literal values.

justinethier:
> > @BrianSipos Addressed comments from your latest review. Please let me know if you have more questions/concerns.
> 
> This is not a code issue but I notice your text ARIs in C comments have bad syntax. The two slash "ari://" is for object reference values not literal values.

Good catch! Alright, that does it for the outstanding issues on this one.

BrianSipos (approved):

PR #52, opened 2024-10-10
ARI updates for object references in text form

justinethier:
Added more test cases for object references.

Fixed an issue where the object type was inconsistently recognized when specified as lower/mixed case, due to a case sensitive `HASH` method.


test/cace/test_ari_text.c
```diff
@@ -895,6 +895,73 @@ void test_ari_text_decode_lit_typed_rptset(const char *text, enum ari_prim_type_
...
+TEST_CASE("ari://test/TD/hi", ARI_TYPE_TD)
+TEST_CASE("ari://test/LABEL/hi", ARI_TYPE_LABEL)
+TEST_CASE("ari://test/CBOR/hi", ARI_TYPE_CBOR)
+TEST_CASE("ari://test/ARITYPE/hi", ARI_TYPE_ARITYPE)
+TEST_CASE("ari://test/AC/hi", ARI_TYPE_AC)
+TEST_CASE("ari://test/AM/hi", ARI_TYPE_AM)
+TEST_CASE("ari://test/TBL/hi", ARI_TYPE_TBL)
+TEST_CASE("ari://test/EXECSET/hi", ARI_TYPE_EXECSET)
+TEST_CASE("ari://test/RPTSET/hi", ARI_TYPE_RPTSET)
+TEST_CASE("ari://test/OBJECT/hi", ARI_TYPE_OBJECT)
```
BrianSipos, line 922: These are valuable test inputs but the correct behavior should actually be to treat these as invalid. While they are all using valid built-in types none of these are valid AMM _object_ types. This was a defect in the original logic, do you mind looking into it? We might need to have more internal flags associated with the built-in type text.

BrianSipos (commented):
The missing hash logic is a good catch and uncovering the unexpected behavior in my earlier comment is also valuable. Let me know if you want me to investigate how to discriminate internally between valid object type names.

test/cace/test_ari_text.c
justinethier: No problem, I'll look into it. But I do want to make sure we're on the same page with these changes. So to confirm, for AMM object types we want to accept the following (from https://github.com/JHUAPL-DTNMA/dtnma-tools/blob/c7ef3dddd6db598385e7b39777d409be2e147308/src/cace/ari/type.h#L69):

```
    ARI_TYPE_IDENT   = -1,
    ARI_TYPE_CONST   = -2,
    ARI_TYPE_CTRL    = -3,
    ARI_TYPE_EDD     = -4,
    ARI_TYPE_OPER    = -6,
    ARI_TYPE_SBR     = -8,
    ARI_TYPE_TBR     = -10,
    ARI_TYPE_VAR     = -11,
    ARI_TYPE_TYPEDEF = -12,
```

And we want to reject the other ARI types, including `ARI_TYPE_OBJECT`, correct?

BrianSipos: Yes exactly, and since these are part of the compiled code a validity check could just be the range of valid (for now) code points with a comment about updating the range if new code points are allocated.

src/cace/ari/base.c
```diff
@@ -215,6 +215,11 @@ void ari_objpath_copy(ari_objpath_t *obj, const ari_objpath_t *src)
     obj->ari_type     = src->ari_type;
 }
 
+bool ari_valid_type_for_objpath(ari_type_t type)
```
BrianSipos, line 218: Very minor feedback; if this is a local-only function (no header declaration) mark it `static` to inform the compiler.

BrianSipos (commented):
I think this is in a very good state except for the static function marking. If you do make that change consider this approved.

src/cace/ari/base.c
justinethier: Done, thanks @BrianSipos!

PR #53, opened 2024-10-16
ARI handling updates for namespace and ARI references

justinethier:
Added additional unit tests for both types of references.

Updated the encoder to emit properly-formatted ARI references. For now do not enforce removal of the scheme but instead leave that to the options passed to the encoder.


BrianSipos (approved):
I made some small additions to cases and this looks good.

justinethier:
Merged - thanks @BrianSipos!

PR #55, opened 2024-10-18
Run formatting job and add it to CI

justinethier:
Bring code formatting up-to-date and add a CI job to check for in going forward.


BrianSipos (approved):
I moved the job into a separate workflow to make it easier to reuse elsewhere but kept the same structure. This is a useful thing to have in CI!

PR #56, opened 2024-10-25
ARI handling updates for references in binary form

justinethier:
Added additional test cases for references in binary form and added new helper functions for constructing objref path ARI objects.

### DIFF
--- a/src/cace/ari/access.h
+++ b/src/cace/ari/access.h
@@ -136,7 +136,8 @@ void ari_set_objref_path_textid(ari_t *ari, const char *ns_id, ari_type_t type_i
  *
  * @param[in,out] ari The ARI value to modify.
  * @param[in] ns_id The namespace path segment, or NULL for none.
- * @param type_id The object type path segment, or NULL for none. The pointed-to lifetime does not need to outlast this function call.
+ * @param type_id The object type path segment, or NULL for none. The pointed-to lifetime does not need to outlast this
+ * function call.
  * @param[in] obj_id The object ID path segment, or NULL for none.
  */
 void ari_set_objref_path_textid_opt(ari_t *ari, const char *ns_id, const ari_type_t *type_id, const char *obj_id);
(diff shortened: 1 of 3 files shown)


src/cace/ari/access.h
```diff
@@ -132,6 +132,18 @@ struct ari_rptset_s *ari_init_rptset(ari_t *ari);
...
+ *
+ * @param[in,out] ari The ARI value to modify.
+ * @param[in] has_ns Determine whether namespace path segment is present.
+ * @param[in] ns_id The namespace path segment.
+ * @param[in] has_type Determine whether object type path segment is present.
+ * @param type_id The object type path segment.
+ * @param[in] has_obj Determine whether object ID path segment is present.
+ * @param[in] obj_id The object ID path segment.
+ */
+void ari_set_objref_path_textid_opt(ari_t *ari, bool has_ns, const char *ns_id, bool has_type, ari_type_t type_id, bool has_obj, const char *obj_id);
```
BrianSipos, line 145: Consider using `const type *` pointers for optional parameters and allowing NULL pointer to indicate lack-of-value. The API doc for the parameter should indicate something like "The pointed-to lifetime does not need to outlast this function call." I can submit a patch commit to give an example if that helps.

src/cace/ari/access.h
```diff
@@ -141,6 +153,18 @@ void ari_set_objref_path_textid(ari_t *ari, const char *ns_id, ari_type_t type_i
...
+ *
+ * @param[in,out] ari The ARI value to modify.
+ * @param has_ns Determine whether namespace path segment is present.
+ * @param ns_id The namespace path segment.
+ * @param has_type Determine whether object type path segment is present.
+ * @param type_id The object type path segment.
+ * @param has_obj Determine whether object ID path segment is present.
+ * @param obj_id The object ID path segment.
+ */
+void ari_set_objref_path_intid_opt(ari_t *ari, bool has_ns, int64_t ns_id, bool has_type, ari_type_t type_id, bool has_obj, int64_t obj_id);
```
BrianSipos, line 166: Same comment as above, to use `const type *` parameter types.

BrianSipos (changes requested):

src/cace/ari/access.h
justinethier: @BrianSipos Please see my most recent changes. For this one and below.

BrianSipos (approved):
Updates look good!

PR #59, opened 2024-11-04
ARI Handling Updates - Literal types in binary form

justinethier:
This PR adds many new test cases for ARI's of literal types in binary form as well as a handful of related changes:

- Added checks to ensure binary-form `TBL` objects contain the "number of columns" parameter as well as full rows of objects.
- Added validation to ensure nonce values are not represented using negative integers. This change applies to both text and binary form ARI's.


BrianSipos (approved):
Very good updates

justinethier:
Thanks @BrianSipos !

PR #61, opened 2024-11-05
ARI handling updates for label types in text form

justinethier:
Added ability to parse numeric label types and added a full set of test cases for text-based `LABEL` types.

### DIFF
--- a/src/cace/ari/text_val.lex
+++ b/src/cace/ari/text_val.lex
@@ -65,6 +65,37 @@ DECDIG [0-9]
     return T_BOOL;
 }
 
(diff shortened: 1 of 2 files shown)


BrianSipos (approved):
These specific changes look good. If the source spec allows negative labels then this might be a separate issue to discuss in that document.

justinethier:
> These specific changes look good. If the source spec allows negative labels then this might be a separate issue to discuss in that document.

We're good on that. The ABNF only allows positive integer labels:

```
id-num = nopad-number

nopad-number = "0" / non-zero-number
non-zero-number = %x31-39 *DIGIT
DIGIT = %x30-39
```

PR #66, opened 2024-11-20
Split CMakeLists files and remove unused deps and old ADM files

BrianSipos:
This is not a change in source, just in build config and removing unused content.


CMakeLists.txt
```diff
@@ -20,7 +20,7 @@ cmake_minimum_required(VERSION 3.10)
 option(BUILD_SHARED_LIBS "Build using shared libraries" ON)
 option(BUILD_DOCS "Enable autodoc building" OFF)
 option(BUILD_AGENT "Build the Agent library and executable" ON)
-option(BUILD_MANAGER "Build the Manager library and executable" ON)
+option(BUILD_MANAGER "Build the Manager library and executable" OFF) # FIXME default to ON
```
justinethier, line 23: **Question**: Do we want to create an issue to track this or is a FIXME good enough?

deps.sh
```diff
@@ -30,7 +30,7 @@ BUILDDIR=${BUILDDIR:-${SELFDIR}/deps/build}
 echo "Building in ${BUILDDIR}"
 echo "Installing to ${DESTDIR}"
 
-if [ ! -e ${DESTDIR}/usr/include/ion.h ]
+if [ false && ! -e ${DESTDIR}/usr/include/ion.h ]
```
justinethier, line 33: **Question**: Is this intentional and do we want a comment explaining why?

CMakeLists.txt
BrianSipos: There will be several tickets after the apl-fy24 merge, updating the manager will be one.

deps.sh
BrianSipos: For the moment this project no longer has a build dependence on ION and it's a slow CI step, but there should be a ticket to re-integrate with ION BP transport after apl-fy24 merge.

CMakeLists.txt
BrianSipos: This is captured in #68

Issue #44, opened 2024-09-12
ARI handling updates to Ref DA


mfarina1:
All updates to the reference DA to use the new ARI handling, must be compliant with the AMM doc section 6 and must be completed roughly by mid October since the Agent implementation is the thing on which many other tools depend (i.e. it’s the API for CAMP and what’s being used by ANMS). 

The goal is for Brian (and Nick if he returns) to do the development and Justin to do the test and cleanup

BrianSipos:
Closed by #70

Issue #25, opened 2024-08-16
Update ADM/ODM objects for latest internet-draft


mfarina1:
Currently unassigned

BrianSipos:
Closed by #70

Issue #16, opened 2024-06-13
handling default manager eid when sending reports


d-linko:
currently setting manager eid to `"dtn:none"` in `lcc.c` when the eid is not given.
ah proposed fix is  agent could have a text-only configuration of its "current manager" within `nmagent_t` that gets set appropriately from `nmagent_register()`

current work around is including manager eid in `rxmgrs` parameter for `gen_rpts`

BrianSipos:
Closed by #70

PR #70, opened 2024-11-21
Base updates to REFDA to enable STDIO Agent use

### LINKED ISSUE #8: Missing an API to inject object handlers
The current Agent object registration API requires a lot of per-ADM generated code and use a form of API that doesn't easily allow for dependency injection. Using an API more similar to NetSNMP with [subtree registration](https://net-snmp.sourceforge.io/dev/agent/group__agent__mib__registering.html) and [object handlers](https://net-snmp.sourceforge.io/dev/agent/group__handler.html#details) would help to separate the Agent behaivior from the object-specific behavior. Some brainstorming can happen in comments on this ticket.

BrianSipos:
This leaves out further updates to the REFDA needed for typing, rule handling, etc. which will be in other specific tickets.

Closes #8 #16 #25 #44


justinethier (approved):
No issues with most recent changes. The older, larger sets of changes have been reviewed in previous PR's.

PR #74, opened 2024-11-25
Add handling of remaining semantic types

BrianSipos:
This refactors other semantic types for consistency.

Closes #58 and #71

PR #77, opened 2024-11-27
Add execution deferral and timeline handling

### LINKED ISSUE #75: Allow CTRL to defer completion
For both waiting controls (described in [Section 4.3.2 of AMM](https://www.ietf.org/archive/id/draft-ietf-dtn-amm-01.html#section-4.3.2)) and app-defined controls, there needs to be some way to indicate that the CTRL callback returning does not mean the completion of CTRL execution.

BrianSipos:
This must be merged after #74.

Closes #75


src/refda/exec.c
```diff
@@ -198,32 +360,77 @@ void *refda_exec_worker(void *arg)
...
+
+            struct timespec nowtime;
+            clock_gettime(CLOCK_REALTIME, &nowtime);
+
+            // execute appropriate callbacks now
+            while (next && timespec_le(next->ts, nowtime))
+            {
+                CACE_LOG_DEBUG("running deferred callback");
+                bool finished = (next->callback)(next->item);
+                if (finished)
```
justinethier, line 392: **Question**: Trying to confirm behavior here. So, if the callback has failed to finish then it will be removed from the execution timeline without the callback being called. If the application isn't done with the item, it has a chance to push the item back on the timeline within `callback`. This cycle could continue indefinitely if necessary.

Is that correct or am I misunderstanding how this API works?

BrianSipos: Somewhat; the timeline is really for internal processing of `ietf-dtnma-agent` controls. Any application can defer finish with a completion callback and will need to "wake" the exec worker with some non-execution sentinel item in the execs queue. This needs to be documented with an example added to unit tests.

justinethier:
Looks good other than the ongoing discussion on completion callbacks. Thanks for putting this together @BrianSipos!

src/refda/exec.c
BrianSipos: This need is addressed by #82 to allow closing this PR now.

PR #81, opened 2024-12-05
Add support for semtype use constraints

### LINKED ISSUE #79: Add support for semtype use constraints
Within the type-use semantic type there should be a container for storing a set of constraints which apply to the semantic type. See ACE implementation for examples.

BrianSipos:
This requires constraints to be added at ADM registration time.

Some constraint types are specifically not included because they affect the validity and visualization that apply outside an Agent context.

This adds upon #77 and must be merged after.

Closes #79


src/cace/amm/range.h
```diff
@@ -0,0 +1,186 @@
...
+            .i_max   = max_val,                                                                          \
+        };                                                                                               \
+    }                                                                                                    \
+                                                                                                         \
+    static inline bool M_C(intvl, _is_infinite)(const M_C(intvl, _t) * obj)                              \
+    {                                                                                                    \
+        return !(obj->has_min) && !(obj->has_max);                                                       \
+    }                                                                                                    \
+                                                                                                         \
+    static inline int M_C(intvl, _cmp)(const M_C(intvl, _t) * lt, const M_C(intvl, _t) * rt)             \
```
justinethier, line 92: **Comment**: The logic in this function is a bit tricky, it's unfortunate that the define prevents adding more detailed comments.

src/cace/amm/semtype.c
```diff
@@ -36,22 +36,37 @@ void amm_semtype_tblt_col_deinit(amm_semtype_tblt_col_t *obj)
...
 static bool amm_semtype_use_match(const amm_type_t *self, const ari_t *ari)
 {
     const amm_semtype_use_t *semtype = self->as_semtype;
     CHKFALSE(semtype);
     const amm_type_t *base = semtype->base;
     CHKFALSE(base);
-    CHKFALSE(base->match);
 
-    if (!base->match(base, ari))
+    if (!amm_type_match(base, ari))
```
justinethier, line 64: **Praise**: Nice upgrade to safer calls here and below.

src/cace/amm/range.h
BrianSipos: Yes, this is a limitation that applies to M*LIB itself also. There still should be extracted doxygen `@struct` level documentation for the macro uses though. So it should be as readable/usable as upstream M*LIB docs.

src/cace/amm/semtype.c
BrianSipos: Yes, any raw uses of function pointers that you notice elsewhere should be avoided.

test/cace/test_amm_semtype_cnst.c
```diff
@@ -0,0 +1,218 @@
...
+    cace_data_deinit(&indata);
+    TEST_ASSERT_EQUAL_INT_MESSAGE(0, res, "ari_cbor_decode() failed");
+
+    res = amm_semtype_cnst_is_valid(cnst, &val);
+    TEST_ASSERT_EQUAL_INT_MESSAGE(expect, res, "amm_semtype_cnst_is_valid() failed");
+
+    ari_deinit(&val);
+}
+
+TEST_CASE("F7", false)             // ari:undefined
```
justinethier, line 59: **Question**: Are these tests using CBOR since that is more likely to be used in fielded systems? Or should we consider using text for unit tests going forward?

test/cace/test_amm_semtype_cnst.c
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    check_cnst(&cnst, inhex, expect);
+    amm_semtype_cnst_deinit(&cnst);
+}
+
+TEST_CASE("[a-z]+", "F7", false)           // ari:undefined
+TEST_CASE("[a-z]+", "F6", false)           // ari:null
+TEST_CASE("[a-z]+", "F4", false)           // ari:false
+TEST_CASE("[a-z]+", "426869", false)       // ari:'hi'
```
justinethier, line 177: **Praise**: Great test case!

test/cace/test_amm_semtype_cnst.c
BrianSipos: Within the C implementation the text decoding is actually optional (depends on FLEX+BISON) so only the binary form is guaranteed to be readable on all CACE configs.

PR #87, opened 2024-12-16
Fix source formatting

BrianSipos:
Also enable CodeQL in normal package environment.


justinethier (approved):
Minimal change, no objections here.

PR #90, opened 2024-12-17
Updates for installation on RHEL 9

justinethier:
Updated `devs.sh` to explicitly pass `-fPIC` to `CFLAGS`. This prevents a linking error when building on RHEL 9.

Updated Readme to include new script / dep in RHEL 9.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,7 @@ civetweb
 libcivetweb-dev
 libssl-dev
 libcjson-dev
+gcovr
 ```
 
 To use the systemd wrapper utility around the executables, also install the PIP package
@@ -55,6 +56,7 @@ systemd-python
 
 Once dependencies are installed, the following scripts can be run to build and test the project:
 ```
+./deps.sh
 ./prep.sh
 ./build.sh check
 ```
--- a/README.md
+++ b/README.md
@@ -54,9 +54,16 @@ To use the systemd wrapper utility around the executables, also install the PIP 
 systemd-python
 ```
 
+In addition to the above, this project also has dependencies provided as git submodules. These may be retrieved and built using the following commands:
+
+```
+git submodule init
+git submodule update --recursive
+./deps.sh
+```
+
 Once dependencies are installed, the following scripts can be run to build and test the project:
 ```
-./deps.sh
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -55,6 +56,7 @@ systemd-python
 
 Once dependencies are installed, the following scripts can be run to build and test the project:
 ```
+./deps.sh
```
BrianSipos, line 59: This command is a little at odds with the text above, maybe separate paragraph about deps (one time thing) from the normal build steps?

deps.sh
```diff
@@ -54,6 +54,7 @@ then
   pushd ${DEPSDIR}/QCBOR
   cmake -S . -B ${BUILDDIR}/QCBOR \
     -DCMAKE_BUILD_TYPE=Debug \
+    -DCMAKE_C_FLAGS="-fPIC " \
```
BrianSipos, line 57: This actually points to maybe a deeper issue in the upstream config. This could be a workaround for us though...

BrianSipos:
I may attempt to add a CI job to build in UBI9 if you don't mind. That will capture this alternate env.

README.md
justinethier: That's a good point, I broke this out separately and also provided `git submodule` instructions.

deps.sh
justinethier: I see what you mean. I was looking at upstream's `Makefile`:

```
CFLAGS=$(CMD_LINE) -I inc -I test -Os -fPIC
```

However they also provide a `CMakeLists.txt` which only calls out `-fPIC` for shared libraries.

justinethier:
> I may attempt to add a CI job to build in UBI9 if you don't mind. That will capture this alternate env.

Much appreciated, thanks! 

The remainder of this PR is ready for re-review when you have a moment.

deps.sh
BrianSipos: It looks like there is some difference in defaults on RHEL, which I made explicit option on the other branch. I'll approve this now and can work it out in the other branch.

BrianSipos (approved):

Issue #85, opened 2024-12-12
Missing elements


jenny-lee-cao:
- [x] Operator desc struct missing operand types
- [x] CACE access missing ari_set_real32

PR #91, opened 2024-12-18
More base adm objects

### LINKED ISSUE #76: Implement all ietf-dtnma-agent objects
This can be assisted with CAMP when it's updated and available.

BrianSipos:
Closes #76 and #85


BrianSipos:
This doesn't fully address #76 but I do want to get this in the dev branch as-is.

justinethier (approved):

PR #92, opened 2024-12-18
Adding job to build and test on RHEL-like environment

BrianSipos:
This also bumps QCBOR to release v1.5 and makes it built explicitly as shared libraries.

### DIFF
--- a/test/refda/test_exec.c
+++ b/test/refda/test_exec.c
@@ -290,7 +290,7 @@ void test_refda_exec_wait_for(const char *targethex, int delay_ms)
     ari_t target = ARI_INIT_UNDEFINED;
     ari_convert(&target, targethex);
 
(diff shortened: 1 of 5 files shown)


test/refda/test_exec.c
```diff
@@ -290,7 +290,7 @@ void test_refda_exec_wait_for(const char *targethex, int delay_ms)
     ari_t target = ARI_INIT_UNDEFINED;
     ari_convert(&target, targethex);
 
-    int wait_ms[] = { delay_ms };
+    int wait_ms[] = { delay_ms, 0 };
```
justinethier, line 293: **Praise**: Good catch!

justinethier (approved):
Looks good, thanks for adding this @BrianSipos

PR #96, opened 2025-01-02
Fixing declarations to allow using CACE and REFDA linked from C++ sources

### LINKED ISSUE #95: Add test targets to compile and link with C++
To verify the ability to use `refda` with a C++ source we should add test executable to link with `cace` and `refda` libraries.

BrianSipos:
Fixing use of cace_amm_obj_desc_t container to avoid implicit strct copies.
Fixing ARI memory/init management.

Closes #95


justinethier (approved):
Approved, changes look good @BrianSipos

PR #97, opened 2025-01-02
Allow building without ARI_TEXT_PARSE

justinethier:
Allow building the project with `ARI_TEXT_PARSE` disabled.


src/cace/cace_ari.c
```diff
@@ -113,13 +113,20 @@ static int read_text(ari_t *inval, FILE *source)
     free(buf);
 
     const char *errm = NULL;
-    res              = ari_text_decode(inval, intext, &errm);
+#if defined(ARI_TEXT_PARSE)
```
BrianSipos, line 116: This leaks memory from `intext` because of this condition. The conditional part could be the entire function body because reading the line will not be useful if the decoder cannot work.

It may be better to actually disallow the in-form to be text entirely, which would avoid the calls to `read_text()`. We can adjust the CACE library API to provide an always-failing `ari_text_decode()` if it makes the compiler logic easier here.

BrianSipos (changes requested):
Some more complicated use of ARI_TEXT_PARSE would be more helpful to the user. Disable the "inform" option for text and the "auto" handling of text form (show an error in that case).

BrianSipos:
This could be made into a proper CI job as well, with an executable-level test of `cace_ari` to verify it works. I think that's currently missing actually.

src/cace/cace_ari.c
justinethier: Good point. Refactored to completely omit `read_text` and `read_auto` if `ARI_TEXT_PARSE` is not defined.

BrianSipos (approved):
I think these all look good and now have good CI test coverage

PR #98, opened 2025-01-04
Use CI job matrix

BrianSipos:
This avoids mostly-duplicate job definitions and makes it easier to test alternate configurations.


justinethier (approved):
Looks good Brian, thanks for simplifying the CI.

PR #99, opened 2025-01-07
Enable refdm and example with sockets

### LINKED ISSUE #68: Organize and update REFDM library from earlier mgr library
This is follow-on to #44 for the manager side.

This also re-enables CMake option `BUILD_MANAGER` by default.

BrianSipos:
This will be draft while specific behaviors are disabled with "FIXME" statements in source.

Addresses part of #68 but does not close.


BrianSipos:
I think this change is big enough for a review and merge.
This contains an end-to-end integration test using AF_UNIX datagram sockets.

README.md
```diff
@@ -112,6 +112,29 @@ A loopback text output can be converted with:
...
+```
+
+The manager can be run from the local installation using:
+```
+./run.sh gdb --args refdm-socket -l debug -a /tmp/mgr.sock
+```
+
+A manual RPTSET can be sent to the manger via command:
+```
+echo -n 821583f61a2d2639a58382250083012205f6 | xxd -r -p | socat STDIO UNIX-SENDTO:/tmp/mgr.sock,bind=/tmp/agent.sock
```
justinethier, line 130: The manager's response to this command is:

```
<ERROR> [msg.c:88:cace_amp_msg_decode] Failed to decode AMP version number, error 6
```

Is that expected?

.github/workflows/codeql.yml
```diff
@@ -67,7 +67,7 @@ jobs:
         sudo apt-get update && sudo apt-get install -y \
           cmake ninja-build \
           ruby build-essential autoconf libtool \
-          flex libfl-dev bison libpcre2-dev civetweb libcivetweb-dev libssl-dev libcjson-dev
+          flex libfl-dev bison libpcre2-dev civetweb libcivetweb-dev libssl-dev libcjson-dev libsystemd-dev
```
justinethier, line 70: **Question**: This dependency doesn't seem necessary to build. Is this required at runtime?

BrianSipos: It is a private build dependency to allow daemons to act nicely under systemd. If not present they will just misbehave when run under systemd.

README.md
BrianSipos: There was a typo at the start of the message content, can you confirm that the fixed version works?

justinethier: This one works much better, thanks @BrianSipos

justinethier (approved):
Reviewed remainder of changes and looks like this one is good to go @BrianSipos

PR #101, opened 2025-01-22
Generalize build scripts

justinethier:
Modifications to the build scripts that allow the reference tools to be more easily integrated with another project.


justinethier:
@BrianSipos this PR contains downstream changes that could be useful to merge back into this repository. Please review when you have a moment.

CMakeLists.txt
```diff
@@ -31,24 +31,28 @@ option(TEST_COVERAGE "Enable test runtime coverage logging" OFF)
...
+  if(NOT GIT_TAG_EXIT EQUAL 0)
+      set(GIT_TAG_REV "v0.0.0")
+      message(WARNING "No git tag found, marking as ${GIT_TAG_REV}")
+  endif()
+  # Make version compatible with CMake needs
+  STRING(REGEX REPLACE [[^v([0-9]+\.[0-9]+\.[0-9]+).*]] [[\1]] GIT_TAG_VERS "${GIT_TAG_REV}")
+  message(STATUS "Using git label ${GIT_TAG_REV} to version ${GIT_TAG_VERS}")
+
+  set(PROJECT_VERSION ${GIT_TAG_VERS})
 endif()
```
BrianSipos, line 52: For simple conditions please add the condition to `endif()` for readability.

BrianSipos (commented):
Was the use of the overloaded [`PROJECT_VERSION`](https://cmake.org/cmake/help/latest/variable/PROJECT_VERSION.html) variable intentional or was that an arbitrary choice?
It's a little confusing to see a normally-set-by-cmake variable name used as control input.

CMakeLists.txt
justinethier: Added a header comment explaining `PROJECT_VERSION` and updated the corresponding `endif` called out above.

BrianSipos (approved):

PR #102, opened 2025-01-23
Enable ION transport binding

### LINKED ISSUE #69: Re-enable ION transport bindings for REFDA and REFDM
This is a follow-on to #44 and #68, and will involve re-enabling ION dep build in `deps.sh` and the `refda-ion` and `refdm-ion` executables.

BrianSipos:
This uses the same send and recv interface functions as earlier `ion_nm_mgr` and `ion_nm_agent` but with the new interface.

Closes #69


src/cace/amp/ion_bp.c
```diff
@@ -0,0 +1,246 @@
...
+
+    if (!retval)
+    {
+        Object newBundle = 0;
+
+        CACE_LOG_DEBUG("sending datagram with %zd octets to %s", msg_size, m_string_get_cstr(meta->dest));
+        int res = bp_send(state->sap,
+                          (char *)m_string_get_cstr(meta->dest), // recipient
+                          NULL,                                  // report-to
+                          300,                                   // lifespan (?)
```
justinethier, line 130: Code itself looks great here, good use of comments.

I am curious though, is the question mark for lifespan due to the value used? Does that need to be revisited?

justinethier (approved):
Looks good @BrianSipos, have a comment for you on BP lifetime.

src/cace/amp/ion_bp.c
BrianSipos: I think the ION calls were carried over from the earlier implementation along with these comments. That is the "lifetime in seconds" field and I will update the comment.

PR #105, opened 2025-01-31
Implement boolean operators

justinethier:
This PR implements boolean operators as part of the IETF DTNMA agent ADM.

In addition `amm_type_set_use_ref_move` is fixed to bind the semtype USE base type.

Finally, `bool-not` would probably make more sense as unary operator but that is a larger change. Thoughts?


BrianSipos:
The bool-not should be unary and if it's not that is a copy-paste typo that can be fixed here.

src/cace/amm/semtype.c
```diff
@@ -94,10 +94,30 @@ int amm_type_set_use_ref_move(amm_type_t *type, ari_t *name)
     amm_semtype_use_init(semtype);
     ari_set_move(&(semtype->name), name);
 
+    int failcnt = 0;
```
BrianSipos, line 97: I think that this setter should _not_ bind the type object based on name. The function docs for these setters should be clear about this if they are not already.

There is separate logic in the agent to do type binding. If there is some issue with that it should be fixed in `refda/binding.c`.

test/refda/test_eval.c
```diff
@@ -267,6 +267,45 @@ TEST_CASE("821181841A000100002302810A", "82060A")
 TEST_CASE("821183820603820605841A000100002501810A", "820612")
 // ari:/AC/(//65536/EDD/2(10),//65536/VAR/1,//65536/OPER/1(10)) -> /VAST/123476
 TEST_CASE("821183841A000100002302810A831A000100002A01841A000100002501810A", "82061A0001E254")
+
+// ari:/AC/(/BOOL/true,/BOOL/false,//ietf-dtnma-agent/OPER/bool-not) -> /BOOL/true
+TEST_CASE("8211838201F58201F48370696574662D64746E6D612D6167656E742568626F6F6C2D6E6F74", "8201F5")
+// ari:/AC/(/BOOL/false,/BOOL/false,//1/OPER/bool-not) -> /BOOL/false
+TEST_CASE("8211838201F48201F483012568626F6F6C2D6E6F74", "8201F4")
+// FIXME: undefined should be falsy per spec, is this responsibility of the convert function or the OPER?
+//// ari:/AC/(/BOOL/false,undefined,//1/OPER/bool-not) -> /BOOL/false
```
BrianSipos, line 276: Undefined is handled specially for type match and convert in that it always gets passed through. These functions should probably have a fall-through for that as well. Does the undefined value trigger the "WARNING" logs above? The warnings could be removed and just leave the result as the default undefined value. That would be a deterministic and understandable behavior.

test/refda/test_eval.c
```diff
@@ -267,6 +267,45 @@ TEST_CASE("821181841A000100002302810A", "82060A")
 TEST_CASE("821183820603820605841A000100002501810A", "820612")
 // ari:/AC/(//65536/EDD/2(10),//65536/VAR/1,//65536/OPER/1(10)) -> /VAST/123476
 TEST_CASE("821183841A000100002302810A831A000100002A01841A000100002501810A", "82061A0001E254")
+
+// ari:/AC/(/BOOL/true,/BOOL/false,//ietf-dtnma-agent/OPER/bool-not) -> /BOOL/true
+TEST_CASE("8211838201F58201F48370696574662D64746E6D612D6167656E742568626F6F6C2D6E6F74", "8201F5")
+// ari:/AC/(/BOOL/false,/BOOL/false,//1/OPER/bool-not) -> /BOOL/false
+TEST_CASE("8211838201F48201F483012568626F6F6C2D6E6F74", "8201F4")
+// FIXME: undefined should be falsy per spec, is this responsibility of the convert function or the OPER?
```
BrianSipos, line 275: This seems like a discrepancy in the AMM, and should be corrected to remove that condition so that undefined value has no boolean conversion.

BrianSipos (commented):
Operators look good, the type binding I think needs removing and maybe fixing elsewhere if needed. I will check on the AMM doc about falsey undefined.

src/cace/amm/semtype.c
justinethier: Looking at this again, maybe it's as simple as calling `refda_binding_semtype_use` from the generated CAMP code.

test/refda/test_eval.c
justinethier: Added a ticket for this - https://gitlab.jhuapl.edu/dtn/dtn-standards/-/issues/176

src/cace/amm/semtype.c
justinethier: OK, updated the FIXME in `binding.c` to handle operands for an `OPER`.

test/refda/test_eval.c
justinethier: Updated per your recommendations.

justinethier:
> The bool-not should be unary and if it's not that is a copy-paste typo that can be fixed here.

Great! Fixed here, let me put in a ticket for the ADMS repo

BrianSipos (approved):
I think this implementation looks good and the tests cover both boolean outputs to make sure the behavior is good.

PR #111, opened 2025-02-21
Add OPER definitions for numeric operations

justinethier:
Adds the vast majority of missing OPER definitions as well as beginning to fill the gaps with EDD's.

Remaining work outside the scope of this PR includes tbl-filter and generalizing the equal / not-equal operations (presently they only support numeric arguments).


src/cace/amm/numeric.c
```diff
@@ -109,3 +110,165 @@ int cace_amm_numeric_promote_type(cace_ari_type_t *result, const cace_ari_t *lef
     }
     return 0;
 }
+
+int cace_numeric_integer_binary_operator(cace_ari_t *result, const cace_ari_t *lt_val, const cace_ari_t *rt_val,
+                                         cace_binop_uvast op_uvast, cace_binop_vast op_vast)
+{
```
BrianSipos, line 116: Any non-static function would be good to guard their inputs against null with `CHKERR1` since they are exported symbols and can be called from anywhere.

src/cace/amm/numeric.c
```diff
@@ -109,3 +110,165 @@ int cace_amm_numeric_promote_type(cace_ari_type_t *result, const cace_ari_t *lef
...
+    cace_ari_deinit(&rt_prom);
+    return retval;
+}
+
+bool cace_numeric_is_zero(const cace_ari_t *val)
+{
+    switch (val->as_lit.prim_type)
+    {
+        case CACE_ARI_PRIM_UINT64:
+            return val->as_lit.value.as_uint64 == 0;
```
BrianSipos, line 261: Don't need both return and break, only one should be present for each branch.

src/refda-stdio/main.c
```diff
@@ -163,6 +163,7 @@ static int stdin_recv(cace_ari_list_t data, cace_amm_msg_if_metadata_t *meta, ca
                     }
 
                     string_t inhex;
+                    string_init(inhex);
```
BrianSipos, line 166: This is a good catch!

src/refda/adm/ietf_dtnma_agent.c
```diff
@@ -353,13 +365,45 @@ static void refda_adm_ietf_dtnma_agent_edd_num_msg_tx(refda_edd_prod_ctx_t *ctx)
...
+    cace_ari_set_uvast(&result, val);
+    refda_edd_prod_ctx_set_result_move(ctx, &result);
     /*
      * +-------------------------------------------------------------------------+
      * |STOP CUSTOM FUNCTION refda_adm_ietf_dtnma_agent_edd_num_msg_tx BODY
      * +-------------------------------------------------------------------------+
      */
 }
 
+// FUTURE:
```
BrianSipos, line 380: We can adjust the ADM and/or counters to capture what is available and remove what is not.
It is actually perfectly valid for an agent to leave the result as the undefined value as a marker of "I do not know this value".

src/refda/adm/ietf_dtnma_agent.c
```diff
@@ -1333,13 +1382,183 @@ static void refda_adm_ietf_dtnma_agent_oper_negate(refda_oper_eval_ctx_t *ctx)
...
+            result.as_lit.value.as_uint64 = val->as_lit.value.as_uint64 * -1;
+            break;
+        case CACE_ARI_PRIM_INT64:
+            result.as_lit.value.as_int64 = val->as_lit.value.as_int64 * -1;
+            break;
+        case CACE_ARI_PRIM_FLOAT64:
+            result.as_lit.value.as_float64 = val->as_lit.value.as_float64 * -1.0;
+            break;
+        default:
+            // leave lit as default undefined
```
BrianSipos, line 1400: In the ADM itself, it is good practice to log warnings about why a result is undefined.
This should also go into the Conventions.md in doxygen or maybe elsewhere in docs.

src/refda/adm/ietf_dtnma_agent.c
```diff
@@ -1333,13 +1382,183 @@ static void refda_adm_ietf_dtnma_agent_oper_negate(refda_oper_eval_ctx_t *ctx)
...
+
+static cace_ari_uvast numeric_lte_uvast(cace_ari_uvast left, cace_ari_uvast right)
+{
+    return left <= right;
+}
+static cace_ari_vast numeric_lte_vast(cace_ari_vast left, cace_ari_vast right)
+{
+    return left <= right;
+}
+static cace_ari_real64 numeric_lte_real64(cace_ari_real64 left, cace_ari_real64 right)
```
BrianSipos, line 1557: These are verbose in C11 but I think a good pattern to use in the absence of any proper language support.

src/refda/adm/ietf_dtnma_agent.c
```diff
@@ -1461,13 +1721,62 @@ static void refda_adm_ietf_dtnma_agent_oper_bit_not(refda_oper_eval_ctx_t *ctx)
...
+    switch (val->as_lit.prim_type)
+    {
+        case CACE_ARI_PRIM_UINT64:
+            result.as_lit.value.as_uint64 = ~val->as_lit.value.as_uint64;
+            break;
+        case CACE_ARI_PRIM_INT64:
+            result.as_lit.value.as_int64 = ~val->as_lit.value.as_int64;
+            break;
+        default:
+            // leave lit as default undefined
```
BrianSipos, line 1736: Log warning about why it is undefined.

test/refda/test_eval.c
```diff
@@ -308,6 +308,70 @@ TEST_CASE("8211838201F48201F48401012568626F6F6C2D786F72", "8201F4")
 TEST_CASE("8211838201F48206008401012568626F6F6C2D786F72", "8201F4")
 // ari:/AC/(/BOOL/true,/TEXTSTR/test,//1/1/OPER/bool-xor) -> /BOOL/false
 TEST_CASE("8211838201F5820A64746573748401012568626F6F6C2D786F72", "8201F4")
+
```
BrianSipos, line 311: Just as pickiness, it would be good to verify that `BYTE` type literal gets handled properly, at least for one of the inputs with a binary operator. I think it will work fine but good to assert in a test case.

test/refda/test_eval.c
```diff
@@ -336,6 +400,10 @@ void test_refda_eval_target_valid(const char *targethex, const char *expectloghe
 
 // ari:/AC/(/BOOL/false,undefined,//1/1/OPER/bool-and) -> undefined
 TEST_CASE("8211838201F4F78401012568626F6F6C2D616E64", "F7")
+// ari:/AC/(/INT/0x0,/UVAST/1,//1/1/OPER/divide) -> undefined
```
BrianSipos, line 403: These are good to have verified here, are there other major conditions that will cause this output? What about trying integer/bitwise operators with at least one REAL32 input?

BrianSipos (commented):
Overall this looks very good, a few small suggestions to catch edge cases and failure test inputs.

src/refda-stdio/main.c
justinethier: Thanks! Was wondering why the app was segfaulting :)

src/refda/adm/ietf_dtnma_agent.c
justinethier: Fixed and updated docs.

src/refda/adm/ietf_dtnma_agent.c
justinethier: Thanks! I had considered using macros instead but the functions, while verbose, seemed more readable/maintainable.

test/refda/test_eval.c
justinethier: Good call, there is another loose end / FIXME called out for REAL32. Let me look into that...

justinethier: OK, added a test case as well as the required real32 conversion function in `typing.c`.

test/refda/test_eval.c
justinethier: That's a good call. So, the initial test case failed because `cace_ari_equal` does not handle the case where the compared ARI's have a different primitive type (BYTE is int from CBOR decoding but uint from eval).

Anyway, this was solved by adding numeric promotion to `cace_ari_equal` which also gives us generalized eq/ne operators in the ADM.

src/refda/adm/ietf_dtnma_agent.c
justinethier: Appreciate your insight on this. If its OK with you I would prefer to leave the stub as-is for now and flesh it out under a different PR.

BrianSipos (approved):

PR #116, opened 2025-03-11
Initial AMM rule support

justinethier:


BrianSipos (commented):
Some design questions in the objects, can be discussed outside of github.

justinethier:
@BrianSipos This latest set of changes addresses your comments. I also added the `enabled` field to the "listing" EDD's in the ADM. It seemed appropriate to have a way to report on rule status as well.

BrianSipos (approved):
Thanks for taking feedback, there will probably be some additional tweaking of object fields when actually used by the exec worker.

PR #117, opened 2025-03-14
TBR rule execution

justinethier:
Draft of changes to allow execution of time based rules.

### DIFF
--- a/src/cace/util/threadset.c
+++ b/src/cace/util/threadset.c
@@ -56,11 +56,6 @@ int cace_threadset_join(cace_threadset_t tset)
     {
         pthread_t thr;
         cace_threadset_pop_back(&thr, tset);
-        if (pthread_cancel(thr))
-        {
-            CACE_LOG_ERR("Unable to cancel pthread %s, errno = %s", "name", strerror(errno));
-            return 2;
-        }
 
         if (pthread_join(thr, NULL))
         {
--- a/src/cace/util/threadset.c
+++ b/src/cace/util/threadset.c
@@ -56,6 +56,12 @@ int cace_threadset_join(cace_threadset_t tset)
     {
         pthread_t thr;
         cace_threadset_pop_back(&thr, tset);
+        if (pthread_cancel(thr))
+        {
+            CACE_LOG_ERR("Unable to cancel pthread %s, errno = %s", "name", strerror(errno));
+            return 2;
+        }
+
         if (pthread_join(thr, NULL))
         {
             CACE_LOG_ERR("Unable to join pthread %s, errno = %s", "name", strerror(errno));
--- a/src/refda/adm/ietf_dtnma_agent.c
+++ b/src/refda/adm/ietf_dtnma_agent.c
@@ -101,7 +101,7 @@ static void refda_adm_ietf_dtnma_agent_ctrl_wait_cond_check(refda_ctrl_exec_ctx_
         {
             // check again in 1s
             refda_timeline_event_t event = {
-                .purpose       = REFDA_TIMELINE_CTRL,
+                .purpose       = REFDA_TIMELINE_EXEC,
                 .ts            = timespec_add(nowtime, timespec_from_ms(1000)),
                 .exec.item     = ctx->item,
                 .exec.callback = refda_adm_ietf_dtnma_agent_ctrl_wait_cond_check,
@@ -1242,7 +1242,7 @@ static void refda_adm_ietf_dtnma_agent_ctrl_wait_for(refda_ctrl_exec_ctx_t *ctx)
     }
 
     refda_timeline_event_t event = {
-        .purpose       = REFDA_TIMELINE_CTRL,
+        .purpose       = REFDA_TIMELINE_EXEC,
         .ts            = timespec_add(nowtime, duration),
         .exec.item     = ctx->item,
         .exec.callback = refda_adm_ietf_dtnma_agent_ctrl_wait_finished,
@@ -1284,7 +1284,7 @@ static void refda_adm_ietf_dtnma_agent_ctrl_wait_until(refda_ctrl_exec_ctx_t *ct
     }
 
     refda_timeline_event_t event = {
-        .purpose       = REFDA_TIMELINE_CTRL,
+        .purpose       = REFDA_TIMELINE_EXEC,
         .ts            = abstime,
         .exec.item     = ctx->item,
         .exec.callback = refda_adm_ietf_dtnma_agent_ctrl_wait_finished,
--- a/src/refda/adm/ietf_dtnma_agent.c
+++ b/src/refda/adm/ietf_dtnma_agent.c
@@ -950,9 +950,10 @@ potential new CTRL's to support rules:
  - enable-rule
  - disable-rule
 
-Since rule object references include the object type, there can be a single discard-rule with a parameter union of /aritype/sbr and /aritype/tbr
-Along those same lines there can be enable-rule and disable-rule controls with the same parameter type.
-In support of all three, there can be a TYPEDEF for a "rule" type that is the union of SBR and TBR references.
+Since rule object references include the object type, there can be a single discard-rule with a parameter union of
+/aritype/sbr and /aritype/tbr Along those same lines there can be enable-rule and disable-rule controls with the same
+parameter type. In support of all three, there can be a TYPEDEF for a "rule" type that is the union of SBR and TBR
+references.
 */
 
 /* Name: sbr-list
--- a/src/refda/adm/ietf_dtnma_agent.c
+++ b/src/refda/adm/ietf_dtnma_agent.c
@@ -942,6 +942,12 @@ static void refda_adm_ietf_dtnma_agent_edd_var_list(refda_edd_prod_ctx_t *ctx)
      */
 }
 
+/* TODO:
+potential new CTRL's:
+ ensure-sbr, discard-sbr, ensure-tbr, discard-tbr
+
+*/
+
 /* Name: sbr-list
  * Description:
  *   A table of SBR within the agent.
--- a/src/refda/adm/ietf_dtnma_agent.c
+++ b/src/refda/adm/ietf_dtnma_agent.c
@@ -943,9 +943,16 @@ static void refda_adm_ietf_dtnma_agent_edd_var_list(refda_edd_prod_ctx_t *ctx)
 }
 
 /* TODO:
-potential new CTRL's:
- ensure-sbr, discard-sbr, ensure-tbr, discard-tbr
+potential new CTRL's to support rules:
 
+ - ensure-rule
+ - discard-rule
+ - enable-rule
+ - disable-rule
+
+Since rule object references include the object type, there can be a single discard-rule with a parameter union of /aritype/sbr and /aritype/tbr
+Along those same lines there can be enable-rule and disable-rule controls with the same parameter type.
+In support of all three, there can be a TYPEDEF for a "rule" type that is the union of SBR and TBR references.
 */
 
 /* Name: sbr-list
--- a/src/refda/adm/ietf_dtnma_agent.c
+++ b/src/refda/adm/ietf_dtnma_agent.c
@@ -101,9 +101,10 @@ static void refda_adm_ietf_dtnma_agent_ctrl_wait_cond_check(refda_ctrl_exec_ctx_
         {
             // check again in 1s
             refda_timeline_event_t event = {
-                .ts       = timespec_add(nowtime, timespec_from_ms(1000)),
-                .item     = ctx->item,
-                .callback = refda_adm_ietf_dtnma_agent_ctrl_wait_cond_check,
+                .purpose       = REFDA_TIMELINE_CTRL,
+                .ts            = timespec_add(nowtime, timespec_from_ms(1000)),
+                .exec.item     = ctx->item,
+                .exec.callback = refda_adm_ietf_dtnma_agent_ctrl_wait_cond_check,
             };
             refda_ctrl_exec_ctx_set_waiting(ctx, &event);
         }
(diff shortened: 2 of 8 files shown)


src/refda/adm/ietf_dtnma_agent.c
```diff
@@ -942,6 +942,12 @@ static void refda_adm_ietf_dtnma_agent_edd_var_list(refda_edd_prod_ctx_t *ctx)
      */
 }
 
+/* TODO:
+potential new CTRL's:
+ ensure-sbr, discard-sbr, ensure-tbr, discard-tbr
```
BrianSipos, line 947: Since rule object references include the object type, there can be a single `discard-rule` with a parameter union of `/aritype/sbr` and `/aritype/tbr` if that makes sense.
Along those same lines there can be `enable-rule` and `disable-rule` controls with the same parameter type.
In support of all three, there can be a TYPEDEF for a "rule" type that is the union of SBR and TBR references.

src/refda/amm/tbr.h
```diff
@@ -65,6 +65,10 @@ typedef struct
      */
     bool enabled;
 
+    /** Current number of executions
```
BrianSipos, line 68: The AMM is unclear about this, but this (maybe) should be clarified as the number since the rule was enabled. Thoughts?

src/refda/exec.c
```diff
@@ -452,3 +452,171 @@ void *refda_exec_worker(void *arg)
...
+{
+    refda_runctx_ptr_t ctxptr;
+    refda_runctx_ptr_init_new(ctxptr);
+
+    if (refda_runctx_from(refda_runctx_ptr_ref(ctxptr), agent, NULL))
+    {
+        return 2;
+    }
+
+    cace_ari_list_t *targets = &(tbr->action.as_lit.value.as_ac->items);
```
BrianSipos, line 469: The action of a rule should be treated as a single execution target; so an AC action would be handled as a MAC value (there are already functions to do just this).

src/refda/exec.c
```diff
@@ -452,3 +452,171 @@ void *refda_exec_worker(void *arg)
...
+}
+
+/** Compute the next scheduled time at which to run the TBR
+ */
+static int refda_exec_tbr_next_scheduled_time(struct timespec *schedtime, const refda_amm_tbr_desc_t *tbr,
+                                              bool starting)
+{
+    if (starting)
+    {
+        if (!tbr->start_time.is_ref && tbr->start_time.as_lit.ari_type == CACE_ARI_TYPE_TP)
```
BrianSipos, line 555: There is a getter in `ari/access.h` for extracting TP and TD values that should be used here.

src/refda/exec.c
```diff
@@ -452,3 +452,171 @@ void *refda_exec_worker(void *arg)
...
+        {
+            struct timespec reltime = tbr->start_time.as_lit.value.as_timespec;
+            if (reltime.tv_nsec == 0 && reltime.tv_sec == 0)
+            {
+                // Rule is always active, start it now
+                clock_gettime(CLOCK_REALTIME, schedtime);
+            }
+            else
+            {
+                // TODO:
```
BrianSipos, line 569: To simplify this, another TBR member field can be added as a `struct timespec` holding this reference absolute time. Then it's up to the object initializer to set the field appropriately.

src/refda/exec.h
```diff
@@ -58,6 +60,14 @@ int refda_exec_run_seq(refda_exec_seq_t *seq);
  */
 void *refda_exec_worker(void *arg);
 
+/**
+ * Begin periodic execution of a time based rule
+ * @param[in] The agent context pointer
+ * @param[in] The rule to execute
+ * @return Non-zero if the rule could not be started
+ */
+int refda_exec_tbr_start(refda_agent_t *agent, refda_amm_tbr_desc_t *tbr);
```
BrianSipos, line 69: This could be renamed as `_enable` and used to control both the enabled state of the TBR and the bookkeeping for enabled rules.

src/refda/exec.c
```diff
@@ -452,3 +452,171 @@ void *refda_exec_worker(void *arg)
...
+        /**
+         * TODO: need to do a few things here
+         * - create a new type of item (or modify existing) such that it can hold either a CTRL or a rule
+         * - then, we pass an instance of that item as part of the event.agent
+         *   - TBD: who owns data pointed to by item?? Maybe it is part of TBR and is held there
+         * - need to support a different type signature for callback. could expand existing
+         *   signature to support either a CTRL or a rule
+         */
+        refda_timeline_event_t event = { .ts   = schedtime,
+                                         .item = (refda_exec_item_t *)tbr, // TODO: this doesn't work, need a container
```
BrianSipos, line 609: Feel free to add an enum field to the timeline event for a "purpose" or similar, with a union of per-purpose data struct. There are examples of this logic in `cace_ari_idseg_t` and some others where C `union` is used in CACE.

BrianSipos (commented):
This is a good start! I added some detail feedback and suggestions.

src/refda/amm/tbr.h
justinethier: Count since rule was enabled seems to make more sense than an absolute count. I've updated the implementation accordingly.

src/refda/exec.c
justinethier: Are you referring to `refda_exec_exp_mac`? Maybe we can simplify by setting up the sequence in `refda_exec_tbr_action` and then calling into the existing MAC function.

src/refda/exec.c
justinethier: Good call! Switched over to these functions.

src/refda/exec.c
justinethier: That's a good idea! Updated accordingly.

src/refda/adm/ietf_dtnma_agent.c
justinethier: Thanks, let me update the comment for now, then come back around to this after the execution code is in a better place, potentially as a follow-on PR.

src/refda/exec.h
justinethier: The original name was `_enable` :). But yeah, renamed and added start up book-keeping for TBR's. This is a good place to unify all of that.

src/refda/exec.c
justinethier: Simplified code per my previous comment.

src/refda/exec.c
justinethier: Added a `purpose` field and moved CTRL / TBR specific fields to structures behind a `union`.

src/cace/util/threadset.c
```diff
@@ -56,6 +56,12 @@ int cace_threadset_join(cace_threadset_t tset)
     {
         pthread_t thr;
         cace_threadset_pop_back(&thr, tset);
+        if (pthread_cancel(thr))
```
BrianSipos, line 59: This will not allow the work thread to clean itself up properly and should not be used. The specific thread function should have a way to properly, cleanly return when the daemon is stopped.

src/refda/amm/tbr.c
```diff
@@ -34,3 +37,11 @@ void refda_amm_tbr_desc_deinit(refda_amm_tbr_desc_t *obj)
     cace_ari_deinit(&(obj->period));
     memset(obj, 0, sizeof(*obj));
 }
+
+bool refda_amm_tbr_desc_reached_max_exec_count(refda_amm_tbr_desc_t *obj)
+{
+    // Max execution count is only applicable if greater than 0,
+    // so ensure if that is the case, that execution count has
+    // not exceeded the limit.
+    return (obj->exec_count > 0 && obj->max_exec_count > 0 && obj->exec_count >= obj->max_exec_count);
```
BrianSipos, line 46: I suggest wrapping sub-conditions in parentheses to make reading them more clear.

src/refda/exec.c
```diff
@@ -375,91 +376,260 @@ void *refda_exec_worker(void *arg)
...
+    }
 
-        refda_timeline_it_t tl_it;
-        refda_timeline_it(tl_it, agent->exec_timeline);
-        if (!refda_timeline_end_p(tl_it))
+    CACE_LOG_INFO("Worker stopped");
+    return NULL;
+}
+
+void refda_exec_worker_iteration(refda_agent_t *agent)
```
BrianSipos, line 386: Good refactoring, but may need to have some return status to know when to exit the thread loop (see earlier comment about thread completion). Some other APIs return `true` to continue looping and `false` when signaled to stop.

src/refda/exec.c
```diff
@@ -375,91 +376,260 @@ void *refda_exec_worker(void *arg)
...
-            sem_wait(&(agent->execs_sem));
+            CACE_LOG_INFO("Got undefined exec, stopping");
+
+            // flush the input queue but keep the daemon running
+            refda_msgdata_t undef;
+            refda_msgdata_init(&undef);
+            refda_msgdata_queue_push_move(agent->rptgs, &undef);
+            sem_post(&(agent->rptgs_sem));
+
+            return;
```
BrianSipos, line 481: This is the loop exit condition.

src/refda/exec.c
```diff
@@ -375,91 +376,260 @@ void *refda_exec_worker(void *arg)
...
+    {
+        tbr->exec_count++;
+
+        // Schedule next execution of the rule
+        //
+        // FUTURE: Consider setup after the TBR's action is fully executed, rather than here.
+        //  Likely is fine either way but if action takes awhile and the period is short
+        //  it could cause problems with overlapping execution. On the other hand if that
+        //  happens the rule is almost certainly too aggressive.
+        refda_exec_schedule_tbr(agent, tbr, false);
```
BrianSipos, line 553: The scheduling should actually occur independently of the execution time, so that if the config is to execute every 2.0 seconds each subsequent start-to-start time should be exactly 2.0 seconds on the timeline.

There is agent execution telemetry available to know if there becomes a backlog of pending sequences.

src/refda/exec.c
```diff
@@ -375,91 +376,260 @@ void *refda_exec_worker(void *arg)
...
+        return;
+    }
+
+    if (refda_amm_tbr_desc_reached_max_exec_count(tbr))
+    {
+        CACE_LOG_INFO("TBR reached maximum execution count");
+        return;
+    }
+
+    if (tbr->action.is_ref || tbr->action.as_lit.ari_type != CACE_ARI_TYPE_AC)
```
BrianSipos, line 532: This is late to catch this. Can this be caught earlier in the process, with an assumption added to the `action` member (in its doxygen comment) that it must be a valid exec target?

src/refda/exec.h
```diff
@@ -54,10 +56,25 @@ int refda_exec_run_seq(refda_exec_seq_t *seq);
 /** Work thread function for the Agent execution manager.
  *
  * @param[in] arg The context ::refda_agent_t pointer.
- * @return Always NULL pointer.
```
BrianSipos, line 57: This should be left in; it is part of the pthread API needs.

src/refda/timeline.h
```diff
@@ -45,6 +43,39 @@ typedef struct refda_timeline_event_s
...
+        REFDA_TIMELINE_TBR,
+        REFDA_TIMELINE_SBR,
+    } purpose;
+    /** Specific time at which the event should occur.
+     */
+    struct timespec ts;
+
+    union
+    {
+        refda_timeline_exec_event_t exec;
```
BrianSipos, line 76: Slight inconsistency in naming between "CTRL" above and "exec" here, not a problem just a suggestion.

BrianSipos (commented):
Overall things look good, some suggestions and questions added

src/refda/exec.c
justinethier: Great catch! Thanks for the assist on this one, just wasn't seeing it during my subsequent read through the code.

src/refda/exec.c
justinethier: Thanks, refactored to take this into account.

src/refda/exec.h
justinethier: Agree, added back.

src/refda/exec.c
justinethier: Yes, updated the code to catch this prior to the rule being enabled, and updated the doxygen comment with your recommendation.

src/refda/exec.c
justinethier: Cleaned up the comment - thanks for clarifying!

src/refda/timeline.h
justinethier: Good point, though. I renamed to "EXEC" since this is consistent internally as well as with the timeline type.

justinethier:
Appreciate your feedback on this one @BrianSipos! 
All of your comments have been addressed and this is ready for one more review.

test/refda/test_exec.c
```diff
@@ -406,3 +406,28 @@ void test_refda_exec_wait_cond(int delay_ms)
 
     cace_ari_deinit(&target);
 }
+
+// ari:/AC/(//65535/10/CTRL/1,//65535/10/CTRL/2), ari:/TD/1, ari:/TD/60
+TEST_CASE("8211828419FFFF0A22018419FFFF0A2202", "820D01", "820D183C", 1, true)
+void test_refda_exec_time_based_rule(const char *actionhex, const char *starthex, const char *periodhex,
```
BrianSipos, line 412: Could you include a test case for both variations of the start time? That would cover the other possible starting condition.

Since the other variation is an absolute time, you could use an additional func parameter to convert the start time to absolute by adding it to the current time.

Is there a way to verify the period in the unit test if it was a shorter duration? Check for two executions, or even just check the timeline after the first execution?

justinethier: Extended the TBR unit tests to include an absolute start time, and to check for multiple iterations in order to test period.

BrianSipos (approved):

Issue #118, opened 2025-03-18
Execution of sequence PID failed warnings on subsequent EXECSET executions


justinethier:
When executing subsequent `EXECSET` commands on the reference agent, the following warning is alternatively raised:

```
2025-02-19T18:17:29.842752 T:400600C6457A0000 <WARNING> [exec.c:346:refda_exec_waiting] execution of sequence PID 2 failed, continuing 
2025-02-19T18:17:29.842753 T:400600C6457A0000 <WARNING> [exec.c:346:refda_exec_waiting] execution of sequence PID 4 failed, continuing 
2025-02-19T18:17:29.842754 T:400600C6457A0000 <WARNING> [exec.c:346:refda_exec_waiting] execution of sequence PID 6 failed, continuing 
2025-02-19T18:17:29.842756 T:400600C6457A0000 <WARNING> [exec.c:346:refda_exec_waiting] execution of sequence PID 9 failed, continuing 
```

This can be reproduced by sending a command to the `refda-stdio`. For example:

```
ari:/EXECSET/n=123;(//ietf/dtnma-agent/CTRL/inspect(//ietf/dtnma-agent/EDD/sw-version))

821482187B8564696574666B64746E6D612D6167656E742267696E7370656374818464696574666B64746E6D612D6167656E74236A73772D76657273696F6E
```

PR #119, opened 2025-03-18
Issue #118 - Safely remove completed exec sequences

justinethier:
Unfortunately, when using `remove` to clean up completed sequences, sequence objects may be relocated internally via [`memmove`](https://github.com/P-p-H-d/mlib/blob/master/m-deque.h#L681), causing a disconnect between the top-level pointers and those maintained by `seq->items`. This leads to problems reported in #118 where a reference to a completed item is still held in memory, leading to the warning message.

This has been fixed by waiting until after iteration completes before popping any completed items off the front of the exec state queue. This safely removes the items while ensuring older items are removed in a timely manner.

### DIFF
--- a/src/refda/exec.c
+++ b/src/refda/exec.c
@@ -280,7 +280,6 @@ static int refda_exec_waiting(refda_agent_t *agent)
         if (atomic_load(&(refda_exec_item_list_front(seq->items)->waiting)))
         {
             // still waiting
-            // FUTURE: is there a better way than busy waiting, especially for a long wait?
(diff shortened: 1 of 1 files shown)


justinethier:
> For the comment about future: the current logic is not a busy wait, the actual waiting is done on the incoming semaphore which can be woken up when a waiting exec item is finished. Maybe remove the comment and place in a ticket if there is concern?

That's a good idea. I removed the comment and broke this out into a separate issue.

PR #122, opened 2025-03-24
SBR rule execution

justinethier:
Added ability to execute SBR's as well as a few general enhancements:

- Consolidate some of the common code between SBR and TBR. Stopped short of removing all redundancies since there are only two types of rules, but we could consolidate further.
- Identify rules in CACE log output
- Update rule-based performance counters

Other thoughts:
- Would it make sense to add an ARI-based ID to identify each rule? Right now the output uniquely identifies them, but it is based on pointer addresses which is not particularly friendly.
- Suspect immediate execution may not be what we want if SBR `min_interval` is zero. Interested in thoughts on this.

### DIFF
--- a/src/refda/amm/sbr.c
+++ b/src/refda/amm/sbr.c
@@ -31,6 +31,7 @@ void refda_amm_sbr_desc_init(refda_amm_sbr_desc_t *obj)
     obj->max_exec_count = 0;
     obj->init_enabled   = true;
     obj->enabled        = true;
+    obj->exec_count     = 0;
 }
 
 void refda_amm_sbr_desc_deinit(refda_amm_sbr_desc_t *obj)
--- a/src/refda/amm/sbr.c
+++ b/src/refda/amm/sbr.c
@@ -44,6 +44,7 @@ void refda_amm_sbr_desc_deinit(refda_amm_sbr_desc_t *obj)
 
 bool refda_amm_sbr_desc_reached_max_exec_count(refda_amm_sbr_desc_t *obj)
 {
+    CHKFALSE(obj);
     // Max execution count is only applicable if greater than 0,
     // so ensure if that is the case, that execution count has
     // not exceeded the limit.
--- a/src/refda/amm/sbr.c
+++ b/src/refda/amm/sbr.c
@@ -1,0 +1,42 @@
+/*
+ * Copyright (c) 2011-2024 The Johns Hopkins University Applied Physics
+ * Laboratory LLC.
+ *
+ * This file is part of the Delay-Tolerant Networking Management
+ * Architecture (DTNMA) Tools package.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "sbr.h"
+
+void refda_amm_sbr_desc_init(refda_amm_sbr_desc_t *obj)
+{
+    obj->action    = CACE_ARI_INIT_UNDEFINED;
+    obj->condition = CACE_ARI_INIT_UNDEFINED;
+
+    obj->min_interval = CACE_ARI_INIT_UNDEFINED;
+    struct timespec interval;
+    interval.tv_nsec = 0;
+    interval.tv_sec  = 0;
+    cace_ari_set_td(&(obj->min_interval), interval);
+
+    obj->max_exec_count = 0;
+    obj->init_enabled   = true;
+    obj->enabled        = true;
+}
+
+void refda_amm_sbr_desc_deinit(refda_amm_sbr_desc_t *obj)
+{
+    cace_ari_deinit(&(obj->action));
+    cace_ari_deinit(&(obj->condition));
+    cace_ari_deinit(&(obj->min_interval));
+    memset(obj, 0, sizeof(*obj));
+}
--- a/src/refda/amm/sbr.c
+++ b/src/refda/amm/sbr.c
@@ -41,3 +41,11 @@ void refda_amm_sbr_desc_deinit(refda_amm_sbr_desc_t *obj)
     cace_ari_deinit(&(obj->min_interval));
     memset(obj, 0, sizeof(*obj));
 }
+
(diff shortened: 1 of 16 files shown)


src/refda/amm/sbr.c
```diff
@@ -40,3 +41,11 @@ void refda_amm_sbr_desc_deinit(refda_amm_sbr_desc_t *obj)
     cace_ari_deinit(&(obj->min_interval));
     memset(obj, 0, sizeof(*obj));
 }
+
+bool refda_amm_sbr_desc_reached_max_exec_count(refda_amm_sbr_desc_t *obj)
+{
```
BrianSipos, line 46: Public functions need to guard their inputs, just a simple `CHKFALSE(obj)` would work.

src/refda/amm/tbr.c
```diff
@@ -34,3 +37,11 @@ void refda_amm_tbr_desc_deinit(refda_amm_tbr_desc_t *obj)
     cace_ari_deinit(&(obj->period));
     memset(obj, 0, sizeof(*obj));
 }
+
+bool refda_amm_tbr_desc_reached_max_exec_count(refda_amm_tbr_desc_t *obj)
+{
```
BrianSipos, line 42: Same guard comment here.

src/refda/exec.c
```diff
@@ -373,93 +375,434 @@ void *refda_exec_worker(void *arg)
...
+
+            if (!refda_exec_rule_action(agent, seq, &(sbr->action)))
+            {
+                sbr->exec_count++;
+                atomic_fetch_add(&agent->instr.num_sbrs_trig, 1);
             }
         }
+    }
+
+    // Always schedule next execution of the rule
```
BrianSipos, line 726: The schedule-next-evaluation should happen at the start of this function so that the evaluation interval is consistent and independent of the complexity of its condition.

src/refda/exec.c
```diff
@@ -373,93 +375,434 @@ void *refda_exec_worker(void *arg)
...
+        }
+    }
+
+    cace_ari_deinit(&ari_res);
+    refda_runctx_deinit(&runctx);
+
+    return res;
+}
+
+/** Begin a single execution of a state based rule
```
BrianSipos, line 684: For consistency in terminology, if this is periodically evaluating the SBR condition then this should not be called "execution" or named "exec". If the AMM doesn't provide specific guidance this could be called "check" or "evaluate" on the rule instance.

src/refda/exec.c
```diff
@@ -373,93 +375,434 @@ void *refda_exec_worker(void *arg)
...
+    refda_exec_seq_t *seq = refda_exec_seq_list_push_back_new(agent->exec_state);
+    seq->pid              = agent->exec_next_pid++;
+
+    // Expand rule and create exec items, CTRLs are run later by exec worker
+    if (!refda_exec_rule_action(agent, seq, &(tbr->action)))
+    {
+        tbr->exec_count++;
+        atomic_fetch_add(&agent->instr.num_tbrs_trig, 1);
+
+        // Schedule next execution of the rule now so time period is accurate
```
BrianSipos, line 547: Same as the SBR comment, schedule the next evaluation at the start of this function so that the interval is consistent and independent of the runtime for this function.

test/refda/test_exec.c
```diff
@@ -406,3 +406,62 @@ void test_refda_exec_wait_cond(int delay_ms)
...
+
+// ari:/AC/(//65535/10/CTRL/1,//65535/10/CTRL/2), ari:undefined, ari:/TD/1
+TEST_CASE("8211828419FFFF0A22018419FFFF0A2202", "F7", "820D01", 1, true, 1)
+// ari:/AC/(//65535/10/CTRL/1,//65535/10/CTRL/2), ari:/AC/(/BOOL/true), ari:/TD/1
+TEST_CASE("8211828419FFFF0A22018419FFFF0A2202", "8211818201F5", "820D01", 1, true, 0)
+// ari:/AC/(//65535/10/CTRL/1,//65535/10/CTRL/2), ari:/AC/(/INT/1), ari:/TD/1
+TEST_CASE("8211828419FFFF0A22018419FFFF0A2202", "821181820401", "820D01", 1, true, 0)
+// ari:/AC/(//65535/10/CTRL/1,//65535/10/CTRL/2), ari:/AC/(/BOOL/false), ari:/TD/1
+TEST_CASE("8211828419FFFF0A22018419FFFF0A2202", "8211818201F4", "820D01", 1, true, 0)
+void test_refda_exec_state_based_rule(const char *actionhex, const char *condhex, const char *min_interval_hex,
```
BrianSipos, line 444: It's more complex, but can a test function be added that evaluates an EDD (defined in the test fixture) so that we can see one rule evaluation that does not execute the action, change the EDD state, and a second eval that does exec?

BrianSipos (commented):
Overall these changes are very good, my comments are in details and test cases.

src/refda/amm/sbr.c
justinethier: Good catch, thanks!

src/refda/exec.c
justinethier: Good idea, the scheduling code has been relocated accordingly.

src/refda/exec.c
justinethier: Updated code accordingly. Note the actual execution takes place later on. Still, this makes scheduling independent of macro expansion and CTRL placement on the timeline.

test/refda/test_exec.c
justinethier: More complex but definitely a good test. Alright, a new test has been added along those lines.

test/refda/test_exec.c
```diff
@@ -55,6 +56,16 @@ static refda_agent_t agent;
 /// Sequence of executions
 static cace_ari_list_t exec_log;
 
+static int edd_backing_value = 0;
```
BrianSipos, line 59: The logic here is good and the test case is great to have, but this is currently not thread safe. If this is replaced with an `atomic_int` instead (and change read/write to use appropriate functions) it will be safe.

BrianSipos (changes requested):

test/refda/test_exec.c
```diff
@@ -55,6 +59,15 @@ static refda_agent_t agent;
 /// Sequence of executions
 static cace_ari_list_t exec_log;
 
+static void test_reporting_edd_one_int(refda_edd_prod_ctx_t *ctx)
+{
+    cace_ari_t result = CACE_ARI_INIT_UNDEFINED;
+    atomic_int val    = atomic_load(&edd_backing_value);
```
BrianSipos, line 65: Sorry for the churn but I think the type of val here should be plain `int`

justinethier: OK, changed `val` to a plain `int`.

BrianSipos (approved):

PR #124, opened 2025-03-31
Add functionality to wake up pending exec items

justinethier:
This PR adds functionality to wake up pending execution items.

### DIFF
--- a/src/refda/exec_item.h
+++ b/src/refda/exec_item.h
@@ -31,10 +31,10 @@ extern "C" {
 // Forward declaration
 typedef struct refda_exec_seq_s refda_exec_seq_t;
 
-#define EXEC_PENDING  0
-#define EXEC_RUNNING  1
-#define EXEC_WAITING  2
-#define EXEC_COMPLETE 3
+#define REFDA_EXEC_PENDING  0
+#define REFDA_EXEC_RUNNING  1
+#define REFDA_EXEC_WAITING  2
+#define REFDA_EXEC_COMPLETE 3
 
 /** Each item in an execution sequence, which corresponds to a
  * dereferenced control.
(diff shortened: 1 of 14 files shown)


src/refda/exec_item.h
```diff
@@ -30,6 +31,11 @@ extern "C" {
 // Forward declaration
 typedef struct refda_exec_seq_s refda_exec_seq_t;
 
+#define REFDA_EXEC_PENDING  0
```
BrianSipos, line 34: Can these be made as a C enum for consistency? Then you don't need to assign explicit integer values.
The atomic state can still be `int` as all C enums are typed as int values.

BrianSipos (commented):
Just one comment about the form of state definitions. Otherwise approved.

src/refda/exec_item.h
justinethier: No problem, these have been refactored into an `enum`.

src/refda/exec_item.h
```diff
@@ -46,20 +55,32 @@ typedef struct
...
+    /**
+     * Indicator if this item is
+     * 1) yet to be completed or paused
+     * 2) waiting on some external event to complete
+     * 3) completed execution but not finished
+     * While waiting this item cannot be executed and will not yet have a
+     * valid #result.
+     *
+     */
+    atomic_int execution_stage;
```
BrianSipos, line 73: Annoying doc feedback:
State enum descriptions should be moved up to the enum declarations, and then this doc be updated to mention something like "Values are one of the ::refda_exec_item_status_t enumerations".

justinethier: Fixed up docstrings

BrianSipos (approved):

PR #135, opened 2025-05-22
applied format and license

d-linko:


BrianSipos (approved):
There is still something funny happening with the centos item-test, possibly related to new SQL changes, causing sig-11s during test that I can't get a stack trace on. It's disabled for now to get the rest of the changes merged in. Will deal with that later.

PR #136, opened 2025-05-23
ADM initialization for rules

justinethier:
Adding code to fully support initialization of rules from a CAMP-generated ADM.

### DIFF
--- a/src/refda/adm/ietf_dtnma_agent.c
+++ b/src/refda/adm/ietf_dtnma_agent.c
@@ -25,6 +25,7 @@
 
 #include "ietf_dtnma_agent.h"
 #include "refda/agent.h"
+#include "refda/exec.h"
 #include "refda/register.h"
 #include "refda/edd_prod_ctx.h"
 #include "refda/ctrl_exec_ctx.h"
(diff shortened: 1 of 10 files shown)


BrianSipos (commented):
The overall changes are good but the binding procedure allows the rules to have parameters. Is this not prohibited by the AMM? If not that is a doc problem too. Some other object types have comments about parameters in the binding function.

justinethier:
> The overall changes are good but the binding procedure allows the rules to have parameters. Is this not prohibited by the AMM? If not that is a doc problem too. Some other object types have comments about parameters in the binding function.

That's a good point. Parameters are prohibited for `CONST` as well, so the latest commit removes them from all three types.

BrianSipos:
> Parameters are prohibited for `CONST` as well, so the latest commit removes them from all three types.

I hate for this churn, but parameters actually should be allowed for CONST and this is a typo/omission from the AMM draft. Can this be added back in? An update to AMM to clarify that both CONST and VAR can have parameters for a specific use is pending.

justinethier:
> > Parameters are prohibited for `CONST` as well, so the latest commit removes them from all three types.
> 
> I hate for this churn, but parameters actually should be allowed for CONST and this is a typo/omission from the AMM draft. Can this be added back in? An update to AMM to clarify that both CONST and VAR can have parameters for a specific use is pending.

That's alright, let's get it right. To avoid issues with refdm formatting changes that are in this PR, I'm going to close this PR and open a new one with the change mentioned above.

PR #142, opened 2025-06-04
Initialization of rules

### LINKED ISSUE #140: Initialization of rules
REFDA needs a new function, separate from the registration and binding, to handle initialization of rules. This will afford additional flexibility for when a persistence subsystem requires the initialization to happen at a time other than startup.

The REFDA function would exist alongside the agent binding function `refda_agent_bindrefs` and be called after binding. It will not be driven by CAMP.

justinethier:
TODO: update all reference applications to call new init function.

Closes #140

### DIFF
--- a/src/refda-ion/main.c
+++ b/src/refda-ion/main.c
@@ -182,6 +182,16 @@ int main(int argc, char *argv[])
         {
             CACE_LOG_INFO("Agent startup completed");
         }
+
+        if (refda_agent_init_objs(&agent))
+        {
+            CACE_LOG_ERR("Agent object initialization failed");
+            retval = 2;
+        }
+        else
+        {
+            CACE_LOG_INFO("Agent object initialization completed");
+        }
     }
 
 #if defined(HAVE_LIBSYSTEMD)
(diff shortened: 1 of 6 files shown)


src/refda-ion/main.c
```diff
@@ -182,6 +182,16 @@ int main(int argc, char *argv[])
         {
             CACE_LOG_INFO("Agent startup completed");
         }
+
+        if (refda_agent_init_objs(&agent))
```
BrianSipos, line 186: If this occurs after `refda_agent_start()` it will need to be thread safe and lock the agent objects while manipulating them. It could also be refactored into agent start and occur before thread spawning. A longer-term behavior will be to only init if persisted object state #7 is not present.

BrianSipos (commented):
Thanks for putting this in. Only one question about thread safety and exposed API.

src/refda-ion/main.c
justinethier: For now, added locking to `refda_agent_init_objs` to guarantee the necessary thread safety.

PR #144, opened 2025-06-11
Add cace_ari_get_byte function

jacksmn1:
Looks like this function might have been missing


justinethier (approved):
Looks straightforward to me and in agreement with the other ARI access functions.

Issue #141, opened 2025-05-29
Improve documentation for newcomers


nobes888:
Enhance documentation to:
- separate required software packages versus auxiliary software packages
- give backdrop to what each step performs in the README.md file
- give guidance to how to perform an intermediate or complete clean up of the workspace

PR #147, opened 2025-06-18
Improve documentation for newcomers (#141)

justinethier:
* Added extra packages used for building

* Added missing build steps

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,8 @@ ruby
 build-essential
 autoconf
 libtool
-flex libfl-dev
+flex 
+libfl-dev
 bison
 libpcre2-dev
 civetweb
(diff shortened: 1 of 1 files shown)


BrianSipos (approved):
These updates look good. Is there value in this PR also giving some guidance about which (sets of) packages are optional and can be conditionally enabled with CMake?

nobes888:
> These updates look good. Is there value in this PR also giving some guidance about which (sets of) packages are optional and can be conditionally enabled with CMake?

I do believe there is value, but for a newcomer the fine details of what is required vs optional (at the start, but later may be required) would be adding too much complexity for this stage.

Having said that adding a section to the README.md file at some point for more advanced users would be useful.

nobes888:
Reopened to allow for it to be merged rather than abandoned.

PR #150, opened 2025-06-20
Adding a namespace built-in type and its logic

BrianSipos:
The code point reflects a pending update to the ARI draft to reduce encoded size.


justinethier:
Thanks for this change @BrianSipos!

Modifications are fine at a glance but there are unit test / CI errors that need to be addressed.

justinethier (approved):

PR #151, opened 2025-06-20
Fix ADM registration issues introduced by #146

BrianSipos:
The other PR #146 didn't catch these with its checks.


justinethier (approved):
Good set of fixes here, thanks!

PR #152, opened 2025-06-23
Fixing references to objects and structuring to avoid bad dependencies

BrianSipos:
This is follow-on to JHUAPL-DTNMA/dtnma-adms#19 updating the modules themselves


justinethier:
@BrianSipos would it be possible for you to run `camp` one more time to incorporate the `ietf-dtnma-agent.yang` ADM changes from this PR - https://github.com/JHUAPL-DTNMA/dtnma-adms/pull/18

Issue #132, opened 2025-05-21
Allow time types to be used with arithmetic operators


BrianSipos:
The TP and TD types should be processed by the arithmetic operators for these conditions:

- [ ] Taking the difference of two TP values, resulting in a TD value
- [ ] Taking the sum (in either order) of a TP and TD value, resulting in a TP value
- [ ] Taking the sum and difference of two TD values, resulting in a TD value
- [ ] Scaling (with multiply and divide) a TD value by a primitive (int or float), resulting in a TD value

Other combinations need to be detected and treated as failure (undefined result) and should be part of the test cases.

PR #155, opened 2025-07-03
Added initial unit tests.

nobes888:
Review of test cases and organization of test cases

### DIFF
--- a/test/refda/test_eval.c
+++ b/test/refda/test_eval.c
@@ -457,3 +457,128 @@ void test_refda_eval_target_failure(const char *targethex, int expect_res)
     cace_ari_deinit(&result);
     cace_ari_deinit(&target);
 }
+
+
+
+// Test difference of 2 TP values
+//
+// --- Difference of 2 TP values ---
+// ari:/AC/(/TP/20000101T001640Z,/TP/20000101T001640Z,//1/1/OPER/sub) -> /TD/0
+TEST_CASE("821183820C1903E8820C1903E88401012563737562", "820D00")
+// ari:/AC/(/TP/20000101T001640Z,/TP/20000101T001600Z,//1/1/OPER/sub) -> /TD/40
+TEST_CASE("821183820C1903E8820C1903C08401012563737562", "820D1828")
+// ari:/AC/(/TP/20000101T001600Z,/TP/20000101T001640Z,//1/1/OPER/sub) -> /TD/-40
+TEST_CASE("821183820C1903C0820C1903E88401012563737562", "820D3827")
+
+// Test addition (in either order) of TP and TD value:
+//
+// --- Difference of 2 TP values ---
+// ari:/AC/(/TD/1000,/TP/20000101T001640Z,//1/1/OPER/add) -> /TP/20000101T003320Z
+TEST_CASE("821183820D1903E8820C1903E88401012563616464", "820C1907D0")
+// ari:/AC/(/TP/20000101T001640Z,/TD/1000,//1/1/OPER/add) -> /TP/20000101T003320Z
+TEST_CASE("821183820C1903E8820D1903E88401012563616464", "820C1907D0")
+
+// TODO: Based on ticket #132, this should be an error...
+///////////////  ari:/AC/(/TP/20000101T001640Z,/TD/1000,//1/1/OPER/sub) -> /TP/20000101T000000Z
+
+// Test addition or subtraction of 2 TD values:
+//
+// --- Addition of 2 TD values ---
+// ari:/AC/(/TD/7000,/TD/5000,//1/1/OPER/add) -> /TD/12000
+TEST_CASE("821183820D191B58820D1913888401012563616464", "820D192EE0")
+// ari:/AC/(/TD/5000,/TD/7000,//1/1/OPER/add) -> /TD/12000
+TEST_CASE("821183820D191388820D191B588401012563616464", "820D192EE0")
+//
+// --- Subtraction of 2 TD values ---
+// ari:/AC/(/TD/7000,/TD/4000,//1/1/OPER/sub) -> /TD/3000
+TEST_CASE("821183820D191B58820D190FA08401012563737562", "820D190BB8")
+// ari:/AC/(/TD/4000,/TD/7000,//1/1/OPER/sub) -> /TD/-3000
+TEST_CASE("821183820D190FA0820D191B588401012563737562", "820D390BB7")
+
+// Test scaling (multiplication / division) of TD with a primitive (int or float):
+//
+// --- Multiply of TD value with primitive (int or float) ---
+// ari:/AC/(/TD/7000,/INT/0,//1/1/OPER/multiply) -> /TD/0
+TEST_CASE("821183820D191B5882040084010125686D756C7469706C79", "820D00")
+// ari:/AC/(/TD/7000,/INT/1,//1/1/OPER/multiply) -> /TD/7000
+TEST_CASE("821183820D191B5882040184010125686D756C7469706C79", "820D191B58")
+// ari:/AC/(/TD/7000,/INT/7,//1/1/OPER/multiply) -> /TD/49000
+TEST_CASE("821183820D191B5882040784010125686D756C7469706C79", "820D19BF68")
+// ari:/AC/(/TD/7000,/REAL32/3.5,//1/1/OPER/multiply) -> /TD/24500
+TEST_CASE("821183820D191B588208F9430084010125686D756C7469706C79", "820D195FB4")
+// ari:/AC/(/TD/7000,/REAL32/INFINITY,//1/1/OPER/multiply) -> undefined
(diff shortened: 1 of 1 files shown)


test/refda/test_eval.c
```diff
@@ -457,3 +457,128 @@ void test_refda_eval_target_failure(const char *targethex, int expect_res)
...
+TEST_CASE("821183820C1903E882041903E884010125686D756C7469706C79", "F7")
+// ari:/AC/(/INT/1000,/TP/20000101T001640Z,//1/1/OPER/multiply) -> undefined
+TEST_CASE("82118382041903E8820C1903E884010125686D756C7469706C79", "F7")
+// ari:/AC/(/TP/20000101T001640Z,/INT/1000,//1/1/OPER/divide) -> undefined
+TEST_CASE("821183820C1903E882041903E88401012566646976696465", "F7")
+// ari:/AC/(/INT/1000,/TP/20000101T001640Z,//1/1/OPER/divide) -> undefined
+TEST_CASE("82118382041903E8820C1903E88401012566646976696465", "F7")
+void test_refda_eval_target_check(const char *targethex, const char *expectloghex)
+{
+    // TODO: This method is mostly duplicate from: test_refda_eval_target_valid()
```
BrianSipos, line 557: If this really is the same body the two can be combined. If the difference is the undefined result value, that can simply be handled by the expected value and that test assertion removed.

nobes888:
Some auxiliary notes:
 * [1] There are 2 test cases that are commented out (L430, L432). I believe the unit tests and the relating logic is proper, however they fail due to how negative fractional TD values are transformed to/from hex. 
 * [2] A unit test is commented out (L437) that perhaps should be supported, but based on my reading of ticket #132 is not supported.
 * [3] In file ietf_dtnma_agent.c L51, the method eqiv_ari_type() is a partial copy of that in the file numeric.c. What are your thoughts?
 * [4] Left printf lines commented out in method timespec_numeric_sub() to aide with debugging of note [1].

src/refda/adm/ietf_dtnma_agent.c
```diff
@@ -43,8 +43,48 @@
 #include <timespec.h>
 /*   STOP CUSTOM INCLUDES HERE  */
 
+#define NANOS_IN_SEC 1000000000
+
 /*   START CUSTOM FUNCTIONS HERE */
 
+static cace_ari_type_t eqiv_ari_type(const cace_ari_lit_t *lit)
```
justinethier, line 50: I suggest removing this function and relocating your newly-added `timespec_` functions to `numeric.c`. These are still numeric operations, they just happen to work with the time-based data types.

src/refda/adm/ietf_dtnma_agent.c
```diff
@@ -299,6 +339,258 @@ static cace_ari_vast bitwise_xor_vast(cace_ari_vast left, cace_ari_vast right)
...
+    is_result_TD |= typeA == CACE_ARI_TYPE_TP && typeB == CACE_ARI_TYPE_TP;
+    is_result_TD |= typeA == CACE_ARI_TYPE_TD && typeB == CACE_ARI_TYPE_TD;
+    if (is_result_TD == false)
+        // Bail if the result will not be a TD
+        return 1;
+
+    struct timespec valueA_TS = valueA->as_lit.value.as_timespec;
+    struct timespec valueB_TS = valueB->as_lit.value.as_timespec;
+
+    //    fprintf(stdout, "\nDEBUG--->   [A]: (%ld sec, %ld nano)   ", valueA_TS.tv_sec, valueA_TS.tv_nsec);
```
justinethier, line 410: It is possible to keep these as `CACE_LOG_DEBUG` statements if they will be useful, otherwise I recommend deleting the dead code.

src/refda/adm/ietf_dtnma_agent.c
```diff
@@ -299,6 +339,258 @@ static cace_ari_vast bitwise_xor_vast(cace_ari_vast left, cace_ari_vast right)
...
+    //    fprintf(stdout, "|   [B]: (%ld sec, %ld nano)", valueB_TS.tv_sec, valueB_TS.tv_nsec);
+
+    // Since we are subtracting invert valueTS_B (so we can do addition instead)
```
justinethier, line 413: **Question** is there an advantage to using addition instead of `timespec_sub`?

test/refda/test_eval.c
```diff
@@ -380,46 +380,138 @@ TEST_CASE("821183820402820402840101256A636F6D706172652D6765", "8201F5")
...
+//
+//-----------------------------------------------------------------------------------------------------------
+// TODO: The 2 unit tests below fail and it may be due to bad parsing of negative fractional TD values.
+// // ari:/AC/(/TD/-2.25,/TD/0.75,//1/1/OPER/sub) -> /TD/3
+// TEST_CASE("821183820D822138AE820D8221184B8401012563737562", "820D03")
+// // ari:/AC/(/TD/3.25,/TD/0.75,//1/1/OPER/sub) -> /TD/-2.5
+// TEST_CASE("821183820D8221190145820D8221184B8401012563737562", "820D82202E")
+//
+//-----------------------------------------------------------------------------------------------------------
+// TODO: Based on ticket #132, this should be an error. Should this eventually be supported?
```
justinethier, line 433: Good catch! I think it should, since as you pointed out subtraction is just addition with a different sign. The case seems valid to me.

src/refda/adm/ietf_dtnma_agent.c
nobes888: An alternative is to make the eqiv_ari_type (global?) visible to this file. If I do that or move the timespec_numeric_*() methods to numeric.c - should I prefix the method name with cace_? You mentioned C does not have namespace support. Also this would create 5 methods with increased visibility versus 1? 

Each method we publicly expose, becomes like a contract (for outside users of this package).

src/refda/adm/ietf_dtnma_agent.c
```diff
@@ -43,8 +43,48 @@
...
+    return CACE_ARI_TYPE_NULL;
+}
+
+/**
+ * Ensure that the specified timespec is normalized. Normalization consists of the absolute value of the timespec nano
+ * field being less than 1 billion.
+ *
+ * @param[inout] target The timespec that is to be normalized.
+ */
+static void timespec_normalize(struct timespec *target)
```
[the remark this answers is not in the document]
nobes888: Will do.

src/refda/adm/ietf_dtnma_agent.c
```diff
@@ -299,6 +339,258 @@ static cace_ari_vast bitwise_xor_vast(cace_ari_vast left, cace_ari_vast right)
...
+            // Adjust nanos to take into account the factional second component
+            result_TS.tv_nsec += (sec_frac * NANOS_IN_SEC);
+            break;
+        default:
+            // This should never happen - perhaps throw an exception instead
+            return 1;
+    }
+
+    // Return undefined value, if scalar is infinite
+    if (fpclassify(scalar_double) == FP_INFINITE)
```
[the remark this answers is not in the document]
nobes888: Good idea will do.

test/refda/test_eval.c
```diff
@@ -380,46 +380,138 @@ TEST_CASE("821183820402820402840101256A636F6D706172652D6765", "8201F5")
...
+// --- Subtraction of 2 TD values ---
+// ari:/AC/(/TD/4000,/TD/7000,//1/1/OPER/sub) -> /TD/3000
+TEST_CASE("821183820D190FA0820D191B588401012563737562", "820D190BB8")
+// ari:/AC/(/TD/7000,/TD/4000,//1/1/OPER/sub) -> /TD/-3000
+TEST_CASE("821183820D191B58820D190FA08401012563737562", "820D390BB7")
+// ari:/AC/(/TD/0.25,/TD/3.75,//1/1/OPER/sub) -> /TD/3.5
+TEST_CASE("821183820D82211819820D82211901778401012563737562", "820D82201823")
+//
+//-----------------------------------------------------------------------------------------------------------
+// TODO: The 2 unit tests below fail and it may be due to bad parsing of negative fractional TD values.
```
justinethier, line 426: I think there's a bug in how the dtnma-tools CBOR code handles negative fractional TD values, as the conversions are different between `cace_ari` and `ace_ari`. Can you put a ticket in for this @nobes888? We can revisit these test cases later on as part of that ticket.

nobes888: Will do. Thanks for doing another check for better confirmation.

src/refda/adm/ietf_dtnma_agent.c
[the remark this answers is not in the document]
nobes888: After offline discussion with @justinethier, this method will be kept in place for now. New commented out unit tests has been added to allow for future testing / correction.

src/refda/adm/ietf_dtnma_agent.c
justinethier: All of the exported functions would need to be prefixed with `cace_`, yes. It seems like the functions are generic enough that they would be appropriate in `numeric`.

PR #159, opened 2025-07-12
Add testing with clang compiler

BrianSipos:
This actually caught and has fixed several issues with ARI decoding and with the `cace_ari_cmp()` and `cace_ari_equal()` comparison functions.


src/cace/cace_ari.c
```diff
@@ -180,6 +181,7 @@ static int read_cborhex(cace_ari_t *inval, FILE *source)
     int    res = getline(&buf, &len, source);
     if (res < 0)
     {
+        free(buf);
```
justinethier, line 184: Great catch on these allocations!

PR #164, opened 2025-07-22
Add REFDA item tests

BrianSipos:
Fixing some report-on behavior
Fixing a reporting crash when the target literal was not an AC value


BrianSipos:
@justinethier are there any other EXECSET tests that you want to add at this time?

Otherwise, there is a new helper function `cace_ari_not_undefined()` which I think is useful for all EDD/CTRL/OPER callbacks to assert preconditions beyond "I have an actual parameter value". Maybe there should be a new "ADM Considerations.md" doxygen source to give this kind of guidance..?

BrianSipos:
I spoke with Leo about potential additions, and the constraint that these tests are limited to the as-built REFDA using only the default ADMs.

PR #168, opened 2025-07-29
Implement remaining ADM rule objects

### LINKED ISSUE #78: Update support for rules
Add back TBR and SBR support with a rule processing thread.

This can be compile-time guarded eventually to disable `ietf-dtnma-agent:rules` feature.

justinethier:
Implements remainder of rule CTRL's in the agent ADM.

Adds helper function to determine if a namespace refers to an ODM.
Adds the ability to mark a rule AMM object as obsolete.
Applies recent fixes to rule ADM objects regarding missing/duplicate parameters.

Changes are tested and ready for review. Note that item-tests cannot be added at this time due to an ADM/ODM handling issue with ACE that will be addressed separately.

Closes #78

### DIFF
--- a/src/refda/adm/ietf_dtnma_agent.c
+++ b/src/refda/adm/ietf_dtnma_agent.c
@@ -2270,30 +2270,30 @@ static void refda_adm_ietf_dtnma_agent_ctrl_ensure_sbr(refda_ctrl_exec_ctx_t *ct
     }
 
     REFDA_AGENT_LOCK(agent, );
-    bool valid = false;
+    bool valid = true, rule_exists = false;
     {
         if (NULL != cace_amm_obj_ns_find_obj_name(odm, CACE_ARI_TYPE_SBR, (const char *)obj_name->ptr))
         {
             // FIXME: update fields on existing SBR in this case, instead of just returning??
             CACE_LOG_INFO("SBR already exists");
-            valid = true;
-            goto ensure_sbr_cleanup;
+            rule_exists = true;
         }
 
         if (NULL != cace_amm_obj_ns_find_obj_enum(odm, CACE_ARI_TYPE_SBR, obj_id))
         {
             // FIXME: same comment as above
             CACE_LOG_INFO("SBR already exists");
-            valid = true;
-            goto ensure_sbr_cleanup;
+            rule_exists = true;
         }
     }
 
+    if (!rule_exists)
     { // For ./SBR/sbr_rule
         cace_amm_obj_desc_t  *obj;
         refda_amm_sbr_desc_t *objdata = CACE_MALLOC(sizeof(refda_amm_sbr_desc_t));
         refda_amm_sbr_desc_init(objdata);
         // action
+        if (valid)
         {
             struct cace_ari_ac_s *action_ac = cace_ari_get_ac((cace_ari_t *)ari_action);
             if (action_ac == NULL)
@@ -2301,11 +2301,15 @@ static void refda_adm_ietf_dtnma_agent_ctrl_ensure_sbr(refda_ctrl_exec_ctx_t *ct
                 CACE_LOG_ERR("Invalid ARI received for action");
                 refda_amm_sbr_desc_deinit(objdata);
                 CACE_FREE(objdata);
-                goto ensure_sbr_cleanup;
+                valid = false;
             }
-            cace_ari_set_copy(&(objdata->action), ari_action);
+            else
+            {
+                cace_ari_set_copy(&(objdata->action), ari_action);
+            }
         }
         // condition
+        if (valid)
         {
             struct cace_ari_ac_s *condition_ac = cace_ari_get_ac((cace_ari_t *)ari_condition);
             if (condition_ac == NULL)
@@ -2313,11 +2317,15 @@ static void refda_adm_ietf_dtnma_agent_ctrl_ensure_sbr(refda_ctrl_exec_ctx_t *ct
                 CACE_LOG_ERR("Invalid ARI received for condition");
                 refda_amm_sbr_desc_deinit(objdata);
                 CACE_FREE(objdata);
-                goto ensure_sbr_cleanup;
+                valid = false;
             }
-            cace_ari_set_copy(&(objdata->condition), ari_condition);
+            else
+            {
+                cace_ari_set_copy(&(objdata->condition), ari_condition);
+            }
         }
         //  min-interval
+        if (valid)
         {
             struct timespec ts;
             if (cace_ari_get_td(ari_min_interval, &ts))
@@ -2325,11 +2333,15 @@ static void refda_adm_ietf_dtnma_agent_ctrl_ensure_sbr(refda_ctrl_exec_ctx_t *ct
                 CACE_LOG_ERR("Invalid ARI received for min interval");
                 refda_amm_sbr_desc_deinit(objdata);
                 CACE_FREE(objdata);
-                goto ensure_sbr_cleanup;
+                valid = false;
             }
-            cace_ari_set_td(&(objdata->min_interval), ts);
+            else
+            {
+                cace_ari_set_td(&(objdata->min_interval), ts);
+            }
         }
         //  init_enabled
+        if (valid)
         {
             cace_ari_bool init_enabled;
             if (cace_ari_get_bool(ari_init_enabled, &init_enabled))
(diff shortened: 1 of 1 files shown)


src/cace/amm/obj_ns.c
```diff
@@ -143,3 +143,15 @@ cace_amm_obj_desc_t *cace_amm_obj_ns_find_obj_enum(const cace_amm_obj_ns_t *ns,
     }
     return *found;
 }
+
+bool cace_amm_obj_ns_is_odm(const cace_amm_obj_ns_t *ns)
+{
+    CHKFALSE(ns);
+
+    if (ns->model_id.has_intenum && ns->model_id.intenum < 0) // FIXME: check if name starts with !
```
BrianSipos, line 151: There is an existing boolean function `m_string_start_with_str_p(name, "!")` to check the name

src/refda/adm/ietf_dtnma_agent.c
```diff
@@ -2218,6 +2223,137 @@ static void refda_adm_ietf_dtnma_agent_ctrl_ensure_sbr(refda_ctrl_exec_ctx_t *ct
...
+
+    refda_agent_t *agent = ctx->runctx->agent;
+    REFDA_AGENT_LOCK(agent, );
+
+    cace_amm_obj_ns_t *odm = cace_amm_obj_store_find_ns(&(agent->objs), odm_ns);
+
+    if (!odm)
+    {
+        CACE_LOG_INFO("ODM not found");
+        return;
```
BrianSipos, line 2247: Returning early like this will both leak the stack vars (`ari_result`) but worse will leave the agent lock. A good pattern followed elsewhere is to have an `bool valid = true`, guard each logical section with `if (valid) { ... }`, and if there are any failures set valid to false. Could use other state similarly, but cannot just return early.

Alternately, these conditions can be checked earlier on before allocating stack and taking agent lock.

src/refda/adm/ietf_dtnma_agent.c
```diff
@@ -2250,6 +2385,151 @@ static void refda_adm_ietf_dtnma_agent_ctrl_ensure_tbr(refda_ctrl_exec_ctx_t *ct
...
+    const cace_ari_t *ari_period       = refda_ctrl_exec_ctx_get_aparam_index(ctx, 5);
+    const cace_ari_t *ari_max_count    = refda_ctrl_exec_ctx_get_aparam_index(ctx, 6);
+    const cace_ari_t *ari_init_enabled = refda_ctrl_exec_ctx_get_aparam_index(ctx, 7);
+
+    cace_ari_t ari_result;
+    cace_ari_init(&ari_result);
+    cace_ari_set_uint(&ari_result, 0);
+
+    refda_agent_t *agent = ctx->runctx->agent;
+    REFDA_AGENT_LOCK(agent, );
```
BrianSipos, line 2402: Same comment about guarding above stack and locking and/or keeping a "valid" running state to avoid bad leaks.

src/refda/adm/ietf_dtnma_agent.c
```diff
@@ -1874,8 +1880,8 @@ static void refda_adm_ietf_dtnma_agent_ctrl_ensure_odm(refda_ctrl_exec_ctx_t *ct
         return;
     }
 
-    org_name = CACE_MALLOC(strlen(org->ptr) + 1);
-    strcpy(org_name, org->ptr);
+    org_name = CACE_MALLOC(strlen((char *)org->ptr) + 1);
+    strcpy(org_name, (char *)org->ptr);
```
BrianSipos, line 1884: Some better bookkeeping of these allocations is needed to avoid opportunity for leaks. The main container can define an object pool for `m_string_t` values for ODM names. Let me know if you want a pointer about doing this.

BrianSipos (changes requested):
The big picture looks good, some specific comments about resource control in C.

src/cace/amm/obj_ns.c
justinethier: Thanks, updated to include this check.

src/refda/adm/ietf_dtnma_agent.c
justinethier: Updated the logic to avoid keeping the lock or leaking memory. Please let me know if you have concerns with the new version.

src/refda/adm/ietf_dtnma_agent.c
justinethier: Updated the code to prevent memory leaks from these allocations within `refda_adm_ietf_dtnma_agent_ctrl_ensure_odm`. Let me know if you still recommend the object pool approach.

src/cace/amm/obj_ns.c
```diff
@@ -143,3 +143,15 @@ cace_amm_obj_desc_t *cace_amm_obj_ns_find_obj_enum(const cace_amm_obj_ns_t *ns,
     }
     return *found;
 }
+
+bool cace_amm_obj_ns_is_odm(const cace_amm_obj_ns_t *ns)
+{
+    CHKFALSE(ns);
+
+    if ((ns->model_id.has_intenum && ns->model_id.intenum < 0) || m_string_start_with_str_p(ns->model_id.name, "!"))
```
BrianSipos, line 151: Recommend to use parentheses around each clause in a condition like this, to be clear to a human (the compiler doesn't care in this case).
```suggestion
    if ((ns->model_id.has_intenum && (ns->model_id.intenum < 0)) || m_string_start_with_str_p(ns->model_id.name, "!"))
```

src/refda/adm/ietf_dtnma_agent.c
```diff
@@ -2218,6 +2225,152 @@ static void refda_adm_ietf_dtnma_agent_ctrl_ensure_sbr(refda_ctrl_exec_ctx_t *ct
...
+
+    cace_ari_int obj_id;
+    if (cace_ari_get_int(ari_obj_enum, &obj_id))
+    {
+        CACE_LOG_ERR("Unable to retrieve object ID");
+        return;
+    }
+
+    {
+        if (NULL != cace_amm_obj_ns_find_obj_name(odm, CACE_ARI_TYPE_SBR, (const char *)obj_name->ptr))
```
BrianSipos, line 2271: Do these accesses also need to lock the agent objects? Consider future situations where these callbacks may be threaded and operate concurrently? The lock should must a critical section where there is cross-logic like "check a thing exists and alter the state". If each part is done in its own critical section it will not be consistent.

src/refda/adm/ietf_dtnma_agent.c
```diff
@@ -2274,6 +2587,77 @@ static void refda_adm_ietf_dtnma_agent_ctrl_ensure_rule_enabled(refda_ctrl_exec_
...
      */
+    const cace_ari_t *target      = refda_ctrl_exec_ctx_get_aparam_index(ctx, 0);
+    const cace_ari_t *ari_enabled = refda_ctrl_exec_ctx_get_aparam_index(ctx, 1);
+
+    cace_ari_t ari_result;
+    cace_ari_init(&ari_result);
+    cace_ari_set_uint(&ari_result, 0);
+
+    cace_ari_bool enabled;
+    if (cace_ari_get_bool(ari_enabled, &enabled))
```
BrianSipos, line 2598: Consider moving this precondition above the allocation of `ari_result` to avoid leak, and adding a check on the content of target (even if the check is just `cace_ari_not_undefined()`). PR #164 adds that function and needs a review. PR #169 also adds a simple check for all parameters not-undefined.

BrianSipos (commented):
I think this is very close, and if there are investigations on lock safety but it's okay for single-thread that can be opened in a separate issue. If you do that, please reference specific functions, lines, and critical sections in that issue.

src/refda/adm/ietf_dtnma_agent.c
```diff
@@ -2258,6 +2407,170 @@ static void refda_adm_ietf_dtnma_agent_ctrl_ensure_tbr(refda_ctrl_exec_ctx_t *ct
...
+
+    REFDA_AGENT_LOCK(agent, );
+    bool valid = false;
+    {
+        if (NULL != cace_amm_obj_ns_find_obj_name(odm, CACE_ARI_TYPE_TBR, (const char *)obj_name->ptr))
+        {
+            // FIXME: update fields on existing TBR in this case, instead of just returning??
+            CACE_LOG_INFO("TBR already exists");
+            valid = true;
+            goto ensure_tbr_cleanup;
```
BrianSipos, line 2461: Strongly avoid goto! Use the valid state to determine whether to continue normal processing blocks and set the state false here and other uses of goto.

src/refda/adm/ietf_dtnma_agent.c
```diff
@@ -2258,6 +2407,170 @@ static void refda_adm_ietf_dtnma_agent_ctrl_ensure_tbr(refda_ctrl_exec_ctx_t *ct
...
+    cace_amm_obj_ns_t *odm = cace_amm_obj_store_find_ns(&(agent->objs), odm_ns);
+    REFDA_AGENT_UNLOCK(agent, );
+
+    if (!odm)
+    {
+        CACE_LOG_INFO("ODM not found");
+        return;
+    }
+
+    if (odm->model_id.intenum >= 0)
```
BrianSipos, line 2433: There is a separate function to check this in a consistent way.

src/refda/adm/ietf_dtnma_agent.c
justinethier: Fair enough, refactored to use `valid` and a straightforward control flow for both `ensure` rule operations.

PR #176, opened 2025-08-05
Update API docs pages

BrianSipos:
This also fixes a REFDM crash when no DB environment is set.


mfarina1 (commented):
Lots of typos to resolve but then it's good to merge

PR #178, opened 2025-08-11
REFDM use DB storage for RPTSET when available

### LINKED ISSUE #128: Have REFDM use DB storage for RPTSET when available
The existing per-process-lifetime storage for received RPTSET values can be used when there is no DB available. But if available the REFDM should use the persistent DB storage and retrieval of RPTSET values.

The RPTSET DB should also record a timestamp of when each RPTSET was received, set the HTTP [Last-Modified](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Last-Modified) header to match the latest row, and allow a query parameter to filter by "since" that time as a lower non-inclusive bound.

### LINKED ISSUE #179: REFDM does not handle bstr-type nonce values in DB
Currently only integer-type nonce values are handled for EXECSET and RPTSET recording to the SQL DB. Other nonce types will actually crash the REFDM because the ARI access is not even guarded.

BrianSipos:
Closes #128 
Closes #179


BrianSipos:
I think these test cases now cover each of the situations that matter for the REFDM

refdb-sql/postgres/Database_Scripts/10-refdb_core_create_database.sql
```diff
@@ -359,15 +359,12 @@ create table if not exists data_model (
 
 
 CREATE TABLE if not exists registered_agents (
-    registered_agents_id serial  NOT NULL ,
-    agent_id_string VARCHAR(128) NOT NULL DEFAULT 'ipn:0.0',
+    registered_agents_id serial  NOT NULL,
```
nobes888, line 362: The renamed field (agent_id_string → agent_endpoint_uri) is  much clearer.

PR #185, opened 2025-08-19
Implement agent ODM-oriented CTRL's

### LINKED ISSUE #171: Implement remaining ietf-dtnma-agent ODM-oriented CTRL's
The following CTRL's need to be implemented:

- ensure-var
- ensure-const
- obsolete-var
- obsolete-const

justinethier:
Add ODM-oriented controls for VAR and CONST.

Closes #171


item-test/test_refda_stdio.py
```diff
@@ -361,4 +361,51 @@ def test_odm(self):
...
+        rptset = self._wait_rptset().value
+
+        self._send_execset('ari:/EXECSET/n=123;(//ietf/dtnma-agent/CTRL/ensure-var(//ietf/!test-model-1,test-var,1,//ietf/amm-semtype/IDENT/type-use(/ARITYPE/int),/AC/(1)))', False)
+        rptset = self._wait_rptset().value
+
+        self._send_execset('ari:/EXECSET/n=123;(//ietf/dtnma-agent/CTRL/inspect(//ietf/dtnma-agent/EDD/var-list(false)))')
+        rptset = self._wait_rptset().value
+        rpt = rptset.reports[0]
+        self.assertEqual(1, len(rpt.items))
+        self.assertEqual(2, rpt.items[0].value.size)
```
BrianSipos, line 384: For the sake of test understanding, it can be good to add a `self.assertIsInstance` just before digging deeply into members-of-members. Not necessarily a test need but helpful for humans to understand what is being asserted.

In this case, asserting that `items[0].value` is a table(?) object.

item-test/test_refda_stdio.py
```diff
@@ -361,4 +361,51 @@ def test_odm(self):
...
-        self.assertEqual(False, rpt.items[0].value[0][5].value); # Confirm rule is disabled
\ No newline at end of file
+        self.assertEqual(False, rpt.items[0].value[0][5].value); # Confirm rule is disabled
+
+
+    def test_odm_var_const(self):
+        self._start()
+
+        LOGGER.setLevel(logging.INFO)
+
```
BrianSipos, line 371: It's useful in some testing to assert preconditions, in this case inspect the `var-list` and see that it is empty to start with. That ensures that what is tested is that `ensure-var` actually adds something.

item-test/test_refda_stdio.py
```diff
@@ -361,4 +361,51 @@ def test_odm(self):
...
+        self._send_execset(
+            'ari:/EXECSET/n=123;(//ietf/dtnma-agent/CTRL/ensure-odm(ietf, 1, !test-model-1, -1))'
+        )
+        rptset = self._wait_rptset().value
+
+        self._send_execset('ari:/EXECSET/n=123;(//ietf/dtnma-agent/CTRL/ensure-var(//ietf/!test-model-1,test-var,1,//ietf/amm-semtype/IDENT/type-use(/ARITYPE/int),/AC/(1)))', False)
```
BrianSipos, line 377: This is really great to see working 👍

item-test/test_refda_stdio.py
```diff
@@ -361,4 +361,51 @@ def test_odm(self):
...
+        self._send_execset('ari:/EXECSET/n=123;(//ietf/dtnma-agent/CTRL/inspect(//ietf/dtnma-agent/EDD/var-list(false)))')
+        rptset = self._wait_rptset().value
+        rpt = rptset.reports[0]
+        self.assertEqual(1, len(rpt.items))
+        self.assertEqual(0, rpt.items[0].value.size)
+
+        self._send_execset('ari:/EXECSET/n=123;(//ietf/dtnma-agent/CTRL/ensure-const(//ietf/!test-model-1,test-const,1,//ietf/amm-semtype/IDENT/type-use(/ARITYPE/int),/AC/(1)))', False)
+        rptset = self._wait_rptset().value
+
+        self._send_execset('ari:/EXECSET/n=123;(//ietf/dtnma-agent/CTRL/inspect(//ietf/dtnma-agent/EDD/const-list(false)))')
```
BrianSipos, line 398: Similar to above, a precondition check that `const-list` starts empty.

src/refda/adm/ietf_dtnma_agent.c
```diff
@@ -2121,6 +2127,98 @@ static void refda_adm_ietf_dtnma_agent_ctrl_ensure_const(refda_ctrl_exec_ctx_t *
...
+
+    if (refda_ctrl_exec_ctx_has_aparam_undefined(ctx))
+    {
+        CACE_LOG_ERR("Invalid parameter, unable to continue");
+        return;
+    }
+
+    REFDA_AGENT_LOCK(agent, );
+    cace_amm_obj_ns_t *odm = cace_amm_obj_store_find_ns(&(agent->objs), ari_namespace);
+    REFDA_AGENT_UNLOCK(agent, );
```
BrianSipos, line 2146: Having two separate critical sections here is somewhat smelly; is there a guarantee that a NS cannot be removed once added? Is it documented in the object store API?

src/refda/amm/const.h
```diff
@@ -45,6 +45,9 @@ typedef struct
      */
     cace_ari_t value;
 
+    /** Determine whether the const has been marked obsolete
+     */
+    bool obsolete;
```
BrianSipos, line 50: Could this marking be moved up to the base object descriptor? Not a critial change needed here, but I think per the AMM that any objects can be marked as obsolete (which would also flow to a CAMP update). Can address this in a separate ticket if you want.

BrianSipos (approved):
The changes as-is look good, but have a few questions and test suggestions.

src/refda/amm/const.h
justinethier: That's a great idea, and I would prefer to address separately. Let me add a ticket.

item-test/test_refda_stdio.py
justinethier: Added preconditions for VAR and CONST.

item-test/test_refda_stdio.py
justinethier: Good point, added checks for `ari.Table`.

src/refda/adm/ietf_dtnma_agent.c
justinethier: Good point, there is a `deinit` in the API and it's not a good design regardless. Updated "ensure" controls to acquire this lock a single time.

PR #190, opened 2025-08-21
Support building as a cmake submodule

jacksmn1:


deps/CMakeLists.txt
```diff
@@ -0,0 +1,63 @@
+cmake_minimum_required(VERSION 3.10)
+
+option(BUILD_SHARED_LIBS "Build using shared libraries" OFF)
```
BrianSipos, line 3: Are these not duplicative of the top-level options?

src/cace/CMakeLists.txt
```diff
@@ -17,9 +17,9 @@ limitations under the License.
 ]]
 
 # AMM and ARI handling library
-configure_file("config.h.in" "config.h")
+configure_file("config.h.in" "${CMAKE_CURRENT_BINARY_DIR}/cace/config.h")
```
BrianSipos, line 20: Why is this relocated? The binary dir is already under ./cace/

src/cace/CMakeLists.txt
```diff
@@ -130,9 +130,14 @@ target_include_directories(
   cace PUBLIC
   $<INSTALL_INTERFACE:$<INSTALL_PREFIX>/include>
 )
+
+get_filename_component(CACE_BINARY_PARENT "${CMAKE_CURRENT_BINARY_DIR}" DIRECTORY)
+
+target_include_directories(cace PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>)
```
BrianSipos, line 136: I don't think this is correct. Any users including CACE headers should name them as 
```c
#include <cace/asdf.h>
```

test/util/CMakeLists.txt
```diff
@@ -28,6 +28,6 @@ add_library(test_util)
 target_sources(test_util PRIVATE ${HFILES})
 target_sources(test_util PRIVATE ${CFILES})
 target_include_directories(
-  test_util PUBLIC
+  test_util PUBLIC ${CMAKE_SOURCE_DIR}/external/dtnma-agent/dtnma-tools/src
```
BrianSipos, line 31: Why is this necessary? The target will inherit the include dirs from the `cace` target itself.

test/CMakeLists.txt
```diff
@@ -17,6 +17,72 @@ limitations under the License.
 ]]
 set(UNITY_ROOT "${CMAKE_SOURCE_DIR}/deps/unity")
 find_package(unitytools)
+set(UNITY_GENERATOR_BIN "${UNITY_ROOT}/auto/generate_test_runner.rb")
```
BrianSipos, line 20: This whole block duplicates the contents of `Findunitytools.cmake`

CMakeLists.txt
```diff
@@ -35,9 +35,10 @@ option(REFDM_UI_CLI "Enable text UI CLI for refdm" OFF)
 option(BUILD_TESTING "Enable test fixtures and libraries" OFF)
 option(TEST_MEMCHECK "Enable test runtime memory checking" OFF)
 option(TEST_COVERAGE "Enable test runtime coverage logging" OFF)
+option(IS_SUBDIRECTORY "Is a subidrectory of another project. Do not build tests." OFF)
```
BrianSipos, line 38: Another PR #159 adds separate option `BUILD_UNITTEST`. Would that be better suited for this purpose?

CMakeLists.txt
```diff
@@ -174,14 +175,44 @@ if(BUILD_TESTING)
...
     --output-on-failure
   )
-endif(BUILD_TESTING)
+endif(BUILD_TESTING AND NOT IS_SUBDIRECTORY)
 
 include(FindPkgConfig)
 
 # Libaries in deps but built separately
-find_package(QCBOR REQUIRED)
+add_subdirectory(deps/QCBOR)
```
BrianSipos, line 183: I don't want to inherit external project build rules in this project. If it is helpful to have in-tree builds of these deps it should be done by its own option. And if QCBOR targets have bad config that needs to be handled with an upstream issue on QCBOR.

CMakeLists.txt
```diff
@@ -174,14 +175,44 @@ if(BUILD_TESTING)
...
+    ${CMAKE_CURRENT_SOURCE_DIR}/deps/QCBOR/inc/qcbor/qcbor_spiffy_decode.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/deps/QCBOR/inc/qcbor/UsefulBuf.h
+)
+set_target_properties(qcbor PROPERTIES
+    PUBLIC_HEADER "${QCBOR_PUBLIC_HEADERS}"
+)
+
 find_package(MLIB REQUIRED)
-find_package(timespec REQUIRED)
+add_subdirectory(deps) # for timespec
```
BrianSipos, line 214: Would really prefer a top-level option to either find_package or subprojects.

BrianSipos (changes requested):
Some changes are okay, some need more discussion.

PR #199, opened 2025-09-08
Avoid memory leak in cace_amp_proxy_cli_send

justinethier:
Ensure `m_bstring_clear` is called in all cases to free allocated memory.


justinethier:
> Looks good.

Thanks Nobes!

PR #201, opened 2025-09-15
Implement agent control flow objects

### LINKED ISSUE #184: Implement ietf-dtnma-agent control flow CTRL's
Implement:

- `if-then-else`
- `catch`

justinethier:
Closes #184

### DIFF
--- a/src/refda/exec.c
+++ b/src/refda/exec.c
@@ -830,14 +830,34 @@ static int refda_exec_action(refda_agent_t *agent, refda_exec_seq_t *seq, const 
 
 int refda_exec_queue(refda_agent_t *agent, const cace_ari_t *ari)
 {
-    CHKVOID(agent);
-    CHKVOID(ari);
+    CHKERR1(agent);
+    CHKERR1(ari);
 
+    int               res = 0;
     refda_exec_seq_t *seq = refda_exec_seq_list_push_back_new(agent->exec_state);
     seq->pid              = agent->exec_next_pid++;
 
-    // Expand target ARI and create exec items, CTRLs are run later by exec worker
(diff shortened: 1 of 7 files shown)


src/refda/exec.c
```diff
@@ -822,3 +805,59 @@ int refda_exec_sbr_disable(refda_agent_t *agent, refda_amm_sbr_desc_t *sbr)
...
+    }
+
+    refda_runctx_ptr_set(seq->runctx, ctxptr);
+    int res = refda_exec_exp_mac(runctx, seq, action);
+
+    refda_runctx_ptr_clear(ctxptr); // Clean up extra reference created by ptr_ref
+    return res;
+}
+
+int refda_exec_queue(refda_agent_t *agent, const cace_ari_t *ari)
```
BrianSipos, line 831: A reason why this implementation was delayed was in trying to integrate the sub-execution (from a target in a CTRL actual parameter) into to main execution pipeline. Is this method here compatible with the pipeline? Meaning can this execution be statused by EDD and possibly interrupted by other CTRL?

BrianSipos (changes requested):
I think this is a great starting point for this behavior and would like to be able to adapt this in a way to integrate with the existing execution pipeline (two-stage expand and run).

Also, this reference behavior should inform the failure handling specified for these CTRLs. Meaning: if one of the actions is not executed, should any bad CTRL references or given parameters in it be ignored at runtime? (the easier behavior is "yes")

src/refda/exec.c
justinethier: @BrianSipos Please let me know your thoughts on the above, and whether it would be better to take the discussion offline.

PR #213, opened 2025-10-06
Fix Manager stopping instead of dropping malformed messages

### LINKED ISSUE #212: Manager ingress worker thread can stop without stopping app
On receipt of an malformed message, the manager's worker thread stops without shutting down the manager.

```
<INFO> [ingress.c:160:refdm_ingress_worker] Got mif.recv result=4, stopping
<INFO> [ingress.c:170:refdm_ingress_worker] Worker stopped
```

Looking at refdm/ingress.c:160, the issue looks to be usage of `break` instead of `continue` in this error handler.

The same may also be present for refda.  The refda version does perform some additional cleanup in this error handling.  The refdm version preserves the same comment but lacks the cleanup -- assuming that flush is not needed for the manager that comment should be revised along with this fix.

DavidEdell:
Fixes #212 behavior to drop malformed messages rather than killing th…e manager's worker thread.

### DIFF
--- a/src/refdm/ingress.c
+++ b/src/refdm/ingress.c
@@ -160,7 +160,7 @@ void *refdm_ingress_worker(void *arg)
             CACE_LOG_INFO("Got mif.recv result=%d, stopping", recv_res);
 
             // flush the input queue but keep the daemon running
-            break;
+            continue;
         }
     }
 
--- a/src/refda/ingress.c
+++ b/src/refda/ingress.c
@@ -79,7 +79,7 @@ void *refda_ingress_worker(void *arg)
             refda_msgdata_queue_push_move(agent->execs, &undef);
             sem_post(&(agent->execs_sem));
 
-            break;
+            continue;
         }
     }
 


src/refdm/ingress.c
```diff
@@ -160,7 +160,7 @@ void *refdm_ingress_worker(void *arg)
             CACE_LOG_INFO("Got mif.recv result=%d, stopping", recv_res);
 
             // flush the input queue but keep the daemon running
```
BrianSipos, line 162: The REFDM doesn't use separate work threads for processing message contents, so this comment can be removed

BrianSipos (changes requested):
I think this behavior is actually correct, and the `recv` function maybe needs better documentation. It should not return if the reception encountered a "normal" or "recoverable" failure, it needs to return with non-zero code to indicate a kind of unrecoverable condition (like the input permanently closed). Any other failures should be caught and handled in the `recv` implementation itself.

PR #223, opened 2025-11-24
IETF DTNMA Agent tbl-filter

### LINKED ISSUE #172: Implement ietf-dtnma-agent tbl-filter
The `tbl-filter` OPER still needs to be implemented in this ADM.

justinethier:
Closes #172

### DIFF
--- a/src/cace/ari/algo.c
+++ b/src/cace/ari/algo.c
@@ -376,16 +376,6 @@ static int cace_ari_translate_ari(cace_ari_t *out, const cace_ari_t *in, const c
     }
     else
     {
-        if (translator->map_lit)
-        {
-            retval = translator->map_lit(&(out->as_lit), &(in->as_lit), ctx);
-            CHKERRVAL(retval);
-        }
-        else
-        {
-            cace_ari_lit_copy(&(out->as_lit), &(in->as_lit));
-        }
-
         if (in->as_lit.has_ari_type)
         {
             switch (in->as_lit.ari_type)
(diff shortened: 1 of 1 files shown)


src/refda/adm/ietf_dtnma_agent.c
```diff
@@ -4127,13 +4125,78 @@ static void refda_adm_ietf_dtnma_agent_oper_compare_le(refda_oper_eval_ctx_t *ct
      */
 }
 
+/**
+ * Helper function to get LABEL data as an integer
+ */
+static int get_label_as_int(cace_ari_t *ari, int *out)
```
BrianSipos: The `access.h` function `cace_ari_get_int()` will do this logic already I think. That will also fail if the value is out-of-range for an INT type.

src/cace/ari/algo.c
```diff
@@ -410,9 +400,30 @@ static int cace_ari_translate_ari(cace_ari_t *out, const cace_ari_t *in, const c
                     break;
                 }
                 default:
+                    if (translator->map_lit)
```
BrianSipos: I think this is the right thing to do here. There is just a lot of nuance to how and when `map_lit` function is used. It may be helpful to head off some confusion and mention in its member doc text that it is not called for any of the container types (AC,AM,TBL).

justinethier: Adding this as a note to the follow-on ticket.

PR #246, opened 2026-02-13
Performing label substitution for CONST and VAR production

### LINKED ISSUE #233: Use LABEL substitution for CONST and VAR production
Current value production from CONST and VAR objects do not expand LABEL values to named/numbered given parameter values.

BrianSipos:
This also adds parameters to `//ietf/dtnma-agent/` ADM to enable defining ODM CONST and VAR objects with formal parameters.

Currently this PR adds upon #245 

Closes #233

### DIFF
--- a/refdb-sql/postgres/Database_Scripts/10-refdb_core_create_database.sql
+++ b/refdb-sql/postgres/Database_Scripts/10-refdb_core_create_database.sql
@@ -501,18 +501,20 @@ create table if not exists ari_tbl (
 );
 
 
--- exec-set
+-- exec-set log
 create table if not exists execution_set(
-    execution_set_id serial not null ,
+    execution_set_id serial not null,
+    -- timestamp from the Manager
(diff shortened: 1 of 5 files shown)


src/refda/adm/ietf_alarms.h
```diff
@@ -0,0 +1,100 @@
+/*
+ * Copyright (c) 2011-2025 The Johns Hopkins University Applied Physics
```
mfarina1: Hopefully this year will be updated by the other merge for updating copyright year

src/refda/alarms.h
```diff
@@ -0,0 +1,259 @@
+/*
+ * Copyright (c) 2011-2025 The Johns Hopkins University Applied Physics
```
mfarina1: Ditto

mfarina1 (commented):
Feel free to resolve all my comments about the year if the other PR takes care of that. Otherwise, everything looks good to me

PR #249, opened 2026-02-14
Allow introspecting ident objects

### LINKED ISSUE #193: Move AMM obsolete fields up to base object descriptor
Brian suggested moving the `obsolete` fields from several AMM objects (VAR, CONST, SBR, TBR) up to the base object descriptor. 

This is a more appropriate location as per the AMM any objects can be marked as obsolete (which would also flow to a CAMP update).

BrianSipos:
This also closes #193

### DIFF
--- a/item-test/run.sh
+++ b/item-test/run.sh
@@ -34,7 +34,7 @@ source ./build/venv/bin/activate
 echo "Installing dependencies..."
 if [[ ! -d ./deps/adms ]]
 then
-    git clone --branch agent-introspect-update https://github.com/JHUAPL-DTNMA/dtnma-adms.git ./deps/adms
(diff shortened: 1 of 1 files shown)


item-test/run.sh
```diff
@@ -34,7 +34,7 @@ source ./build/venv/bin/activate
 echo "Installing dependencies..."
 if [[ ! -d ./deps/adms ]]
 then
-    git clone --branch main https://github.com/JHUAPL-DTNMA/dtnma-adms.git ./deps/adms
+    git clone --branch agent-introspect-update https://github.com/JHUAPL-DTNMA/dtnma-adms.git ./deps/adms
```
BrianSipos: This will need to be put back after ADMs PR is merged

PR #251, opened 2026-02-18
nm_sql.c cleanup for improved concurrency, removal of deprecated MySQL code

### LINKED ISSUE #244: Concurrent access to pgsql DB by multiple threads leads to lockups and other failures
I am investigating an issue where some commands periodically fail.  In the course of this testing, I performed a stress test that seems to result in dtnma tools being unable to perform additional SQL queries.

To reproduce:
- Issue a PUT command to send an ARI and verify it works
- Execute the command several times in rapid succession
- Confirm that it's ability to send commands is locked up, even as other NM commands that do not invoke the DB work as expected

Logging confirms it is hanging in nm_rest.c between lines 475 and 485.  civetweb creates a new task for each connection, so this doesn't affect other API behavior, but could prevent other activities (ie: logging received messages) from working.  

The postgres library does not support multiple threada accessing the same connection simultaneously.  It looks like we currently have a single connection used for all REST API operations, but civetweb will create a new task for each HTTP connection. That opens the probability of conflicting access for long-running queries or a busy system.

The solution is to either establish a new connection for each REST Connection (not recommended), add an additional mutex to protect access to the connection (quick solution), or (preferable) define a connection pool that can be reused.

Note: Connection pooling is that standard procedure in general, but is not a feature provided by the C library.

DavidEdell:
Updated nm_sql.c to consistently protect usage of DB connections agai…nst concurrent access which can result in unpredictable behabior. This commit also simplifies the code for readability by removing the currently unused MySQL support.

This resolved #244.

Setting this as a draft pending additional testing.

### DIFF
--- a/src/refdm/nm_sql.c
+++ b/src/refdm/nm_sql.c
@@ -63,14 +63,14 @@ typedef enum db_con_e
 
 typedef struct refdm_db_pool_t
 {
-    /// Lock for this struct instance and its pointers
+    /// Lock for certain uses of #conn which are not thread safe.
     pthread_mutex_t lock;
     /** Connection handle.
      * This can be used outside of #lock in a thread safe way, but not reassigned.
      */
     PGconn *conn;
     /** Parameters used to connect, which must outlive this pointer.
-     * These parameters are used in case reconnection is needed.
+     * These parameters are used in case re-connection is needed.
      */
     const refdm_db_t *parms;
 } refdm_db_pool_t;


src/refdm/nm_sql.c
```diff
@@ -62,24 +61,22 @@ typedef enum db_con_e
...
+
+/* Global connections to the MYSQL Server. */
+refdm_db_pool_t dbpool[MGR_NUM_SQL_CONNECTIONS];
 
-static pthread_mutex_t db_rest_con_use;
+#define checkConn(idx) (idx < MGR_NUM_SQL_CONNECTIONS && dbpool[idx].conn != NULL)
+#define getConn(idx) pthread_mutex_lock(&dbpool[idx].lock)
+#define giveConn(idx) pthread_mutex_unlock(&dbpool[idx].lock)
+
+static refdm_db_t *gParms;
```
BrianSipos: It looks like this pointer really belongs associated with each connection in `refdm_db_pool_t` since its use elsewhere seems to be along with each connection.

PR #265, opened 2026-04-03
Background execute catch sync try target

### LINKED ISSUE #215: Proper control flow for ietf-dtnma-agent/catch
`catch` currently only checks to see if try was setup properly. It needs to be extended to wait until the try sequence is completed and then call into failure if necessary. 

This implies future logic must be guaranteed to run regardless of success/failure, and provide that status to the callback.

From @BrianSipos :

> The API will be something like: within the sequence state a whole-sequence success/failure indication and a semaphore/condition variable indicating when the sequence has finished. And some way to inject these from the ingress API so that it will only ever be present for internally-generated sequences, external EXECSET will never have or need these.
> 
> Then the catch CTRL will be able to use this same internal API to spawn off the try target, wait for its completion async, and then if it fails run the catch target.

BrianSipos:
Closes #215


BrianSipos:
This depends on #263 and #264 to get tests green.

src/refda/exec_item.c
```diff
@@ -22,6 +22,7 @@
 #include "agent.h"
 #include "runctx.h"
 
+// FIXME: is this needed?? A post is done now in refda_ctrl_exec_ctx_check_result()
```
BrianSipos: Need to check with users

mfarina1 (approved):

PR #271, opened 2026-04-16
Big refactoring of arithmetic and conditional operators

BrianSipos:
These now use consistent type-promotion logic and there are item test cases to cover different varieties of comparisons.

This also adjusts internal execution sequence bookkeeping for object and thread safety.

Follow-on to https://github.com/JHUAPL-DTNMA/dtnma-adms/issues/44


item-test/run.sh
```diff
@@ -37,11 +37,11 @@ do_prep() {
     echo "Installing dependencies..."
     if [[ ! -d ./deps/adms ]]
     then
-        git clone --branch main https://github.com/JHUAPL-DTNMA/dtnma-adms.git ./deps/adms
+        git clone --branch 44-numeric-operators https://github.com/JHUAPL-DTNMA/dtnma-adms.git ./deps/adms
```
justinethier: Noting branches to update here prior to merge

src/cace/ari/text_val.lex
BrianSipos, line 290: This should also limit to 9 digits below point.

src/cace/amm/promote.c
```diff
@@ -0,0 +1,281 @@
...
+
+    if (lt_val->is_ref != rt_val->is_ref)
+    {
+        // no possible common type
+        return 2;
+    }
+
+    if (lt_val->is_ref)
+    {
+        return cace_amm_promote_valtype(result, lt_val);
```
justinethier, line 178: Is it good enough to only look at the left val, or do we need to consider both left and right here

src/cace/amm/promote.c
```diff
@@ -0,0 +1,281 @@
...
+        case CACE_ARI_TYPE_REAL32:
+            return 5;
+        case CACE_ARI_TYPE_REAL64:
+            return 6;
+        default:
+            return -1;
+    }
+}
+
+static int cace_amm_promote_type(cace_ari_type_t *result, const cace_ari_t *lt_val, const cace_ari_t *rt_val)
```
justinethier, line 164: There's a bit more going on with this function could be helpful to document

src/cace/amm/promote.c
BrianSipos: Yes; the immediate previous check is that they are both the same state (ref or not).

src/cace/amm/promote.c
BrianSipos: This implements an AMM procedure, will add docstring about this.

src/cace/amm/promote.h
```diff
@@ -0,0 +1,79 @@
...
+/** Determine if given type is numeric.
+ *
+ * @param[in] typ Input type.
+ * @return true if numeric type.
+ */
+bool cace_amm_promote_is_numeric(cace_ari_type_t typ);
+
+typedef struct
+{
+    const cace_ari_t *lt_val;
```
justinethier, line 47: These 2 are straightforward but should still have a docstring

src/cace/amm/promote.h
```diff
@@ -0,0 +1,79 @@
...
+    /// Optionally used convert storage
+    cace_ari_t rt_prom;
+
+    /// The final left value to use
+    const cace_ari_t *lt_use;
+    /// The final right value to use
+    const cace_ari_t *rt_use;
+} cace_amm_promote_state_t;
+
+void cace_amm_promote_init(cace_amm_promote_state_t *obj);
```
justinethier, line 68: Consider documenting `init`. `deinit` deserves a brief docstring reminding the caller that value members are not affected.

justinethier (approved):

PR #273, opened 2026-04-20
Implement exec-deadline and termination logic

### LINKED ISSUE #161: Resolve defect with how the dntma-tools CBOR code handles negative fractional TD values
Notes:
- Discovered while working on  pull #155 , there appears to be a defect with how negative fractional TD values are transformed to/from hex.
- Also @justinethier confirmed that the conversions are different between cace_ari and ace_ari

### LINKED ISSUE #272: Add flow ctrl for enforcing a timeout
For the builtin `wait-cond` control (and any application-defined long running control) there is no way for a Manager to influence a timeout interval for some execution target (single ctrl or a macro).

Also, the `wait-cond` could use an optional second parameter as a hint for check time interval.

BrianSipos:
This implements the control from https://github.com/JHUAPL-DTNMA/dtnma-adms/pull/57

This also fixes issues with negative decimal fraction values and adds TP and TD encode/decode test cases at domain limits.


src/cace/ari/text_util.c
```diff
@@ -477,8 +477,17 @@ int cace_decfrac_encode(m_string_t out, const struct timespec *in)
 {
     CHKERR1(in);
 
-    m_string_cat_printf(out, "%" PRId64, in->tv_sec);
-    cace_subsec_encode(out, in->tv_nsec);
+    time_t            fullsec = in->tv_sec;
+    cace_ari_subsec_t subsec  = in->tv_nsec;
+    if ((fullsec < 0) && (subsec > 0))
+    {
+        // add subsec to negative value, move closer to zero
```
justinethier, line 484: Logic is sound here, weird edge case.

src/cace/ari/text_util.c
```diff
@@ -513,7 +522,14 @@ int cace_decfrac_decode(struct timespec *out, const cace_data_t *in)
         return 3;
     }
 
-    out->tv_sec  = sec_num;
+    if ((fullsec < 0) && (subsec > 0))
```
justinethier, line 525: This negative/positive logic is in a few places across the file and a bit tricky, would be a good candidate for a helper function.

src/refda/adm/ietf_dtnma_agent.c
```diff
@@ -258,6 +270,89 @@ static void refda_adm_ietf_dtnma_agent_ctrl_catch_finished(bool failed, void *us
...
+{
+    /// Execution status future
+    refda_exec_status_t status;
+    /// Target if status indicates failure
+    cace_ari_t on_timeout;
+    /** Execution item of catch to mark result on, not part of the try target.
+     * This is not reference counted because it is part of the item itself!
+     */
+    refda_exec_item_ptr_t *item_ptr;
+} refda_exec_deadline_data_t;
```
justinethier, line 284: Very similar to the try-catch internal structure. I'm fine with keeping them separate but consider a consolidation if this pattern becomes more widespread.

src/refda/adm/ietf_dtnma_agent.c
```diff
@@ -2412,28 +2507,105 @@ static void refda_adm_ietf_dtnma_agent_ctrl_catch(refda_ctrl_exec_ctx_t *ctx)
...
+        CACE_LOG_ERR("Invalid parameter, unable to continue");
+        return;
+    }
+    const cace_ari_t *target     = refda_ctrl_exec_ctx_get_aparam_index(ctx, 0);
+    const cace_ari_t *deadline   = refda_ctrl_exec_ctx_get_aparam_index(ctx, 1);
+    const cace_ari_t *on_timeout = refda_ctrl_exec_ctx_get_aparam_index(ctx, 2);
+
+    refda_exec_deadline_data_t *state = CACE_MALLOC(sizeof(refda_exec_deadline_data_t));
+    refda_exec_deadline_data_init(state, ctx->item_ptr, on_timeout);
+    // free this in when execution item is finished
```
justinethier: **Nitpick** comment could be clearer.

justinethier (approved):

src/cace/ari/text_util.c
BrianSipos: There is a helper in `timespec_normalize()` but that handles other cases not needed here and this is also really 3 lines.

src/refda/adm/ietf_dtnma_agent.c
BrianSipos: Yes I agree, though each has its own parameters. The user_data member on exec item is commonized though, and is now usable for other applications if needed.

Issue #220, opened 2025-11-04
Enforce base Agent permissions


BrianSipos:
As a follow-on to the ACL bookkeeping from #207 the REFDA needs to query and enforce the other in-built Agent permissions:
```
execute 
produce
modify-var
create-odm
delete-odm
create-object
delete-object
```

PR #277, opened 2026-04-21
Agent ACL enforcement

BrianSipos:
Handles part of #220


src/refda/adm/ietf_dtnma_agent.c
```diff
@@ -4377,9 +4516,23 @@ static void refda_adm_ietf_dtnma_agent_ctrl_obsolete_rule(refda_ctrl_exec_ctx_t
     else if (deref.obj_type == CACE_ARI_TYPE_TBR)
     {
         refda_amm_tbr_desc_t *tbr = deref.obj->app_data.ptr;
-        // FIXME need agent access control
-        CACE_LOG_DEBUG("Marking TBR as obsolete");
-        deref.obj->status = CACE_AMM_STATUS_OBSOLETE;
+
+        // access check, this permission has no parameters
```
justinethier, line 4520: This pattern is repetitive, consider if there is a way to simplify across this file.

justinethier (approved):

src/refda/adm/ietf_dtnma_agent.c
BrianSipos: This actually is simplified in follow-on https://github.com/JHUAPL-DTNMA/dtnma-tools/pull/278 down to one-line function call

PR #278, opened 2026-04-22
Implement reference predicates and ACL target filter

### LINKED ISSUE #220: Enforce base Agent permissions

### LINKED ISSUE #283: Ensure LABEL-substituted reference values are kept as reference
The translate filters that substitute LABEL should replace by `//ietf/dtnma-agent/oper/ref(value)` and the `ref` operator should be made to pass-through literal values unchanged.

_Originally posted by @justinethier in https://github.com/JHUAPL-DTNMA/dtnma-tools/pull/278#discussion_r3174028714_

BrianSipos:
This also refactors common filter processing into `refda_eval_filter()`.

This is a follow-on to #277 and https://github.com/JHUAPL-DTNMA/dtnma-adms/pull/62

Closes implementation of https://github.com/JHUAPL-DTNMA/dtnma-adms/issues/60, closes #220, closes #283


BrianSipos (commented):
Need to implement is-same- operators

BrianSipos:
Item tests passing depends on https://github.com/JHUAPL-DTNMA/dtnma-ace/pull/125 and https://github.com/JHUAPL-DTNMA/dtnma-adms/pull/61

BrianSipos:
There are still FIXME for ACL about modifying the enabled state of rules. This doesn't correlated to any existing permission though.

@justinethier does it seem like a general permission like `modify-rule` or something more specific like `modify-rule-enabled` is appropriate?

BrianSipos:
Item tests depend on https://github.com/JHUAPL-DTNMA/dtnma-ace/pull/125 to pass

src/refda/eval.c
```diff
@@ -290,12 +290,61 @@ int refda_eval_target(refda_runctx_t *runctx, cace_ari_t *result, const cace_ari
...
+        CACE_LOG_ERR("Unable to translate filter, error %d", res);
+        cace_ari_deinit(&sub_tgt); // No longer needed at this point
+        return 2;
+    }
+
+    refda_eval_ctx_t evalctx;
+    refda_eval_ctx_init(&evalctx, runctx);
+
+    // mutex-serialize object store access
+    REFDA_AGENT_LOCK(runctx->agent, 1);
```
justinethier, line 371: Will this be a problem if the caller already holds the mutex. For example var-reset (line 2975) then refda_acl_search_one_permission (line 2993).

src/refda/eval.c
```diff
@@ -290,12 +290,61 @@ int refda_eval_target(refda_runctx_t *runctx, cace_ari_t *result, const cace_ari
...
     }
 
     refda_eval_ctx_deinit(&ctx);
     return res;
 }
+
+int refda_eval_filter(refda_runctx_t *runctx, cace_ari_t *result, const cace_ari_t *target,
+                      const cace_ari_translator_t *translator, void *user_data)
+{
+    cace_ari_t sub_tgt = CACE_ARI_INIT_UNDEFINED;
```
justinethier: Thoughts on the following report:

Reference-valued label substitutions are not preserved as values. refda_eval_filter substitutes labels into sub_tgt at eval.c (line 319), then evaluates the substituted tree normally at eval.c (line 332). Any substituted object reference is then treated as expression syntax by refda_eval_expand_item (line 32), so CONST/VAR/EDD refs are produced, OPER refs are executed, TYPEDEF refs are not pushed as values, and IDENT refs fail. This breaks direct filters like (/label/0, match-object-type(...)) for ACL target refs from acl.c (line 218), alarm resource/category refs, table cells containing refs, and unary-eval ref operands at ietf_dtnma_agent.c (line 6387). The substituted value likely needs to be inserted as an eval value, not re-expanded as expression text, or reference substitutions need contextual quoting via oper/ref.

src/refda/eval.c
justinethier: Looking at this again maybe it's fine since the mutex is created as `PTHREAD_MUTEX_RECURSIVE`.

justinethier (approved):

src/refda/eval.c
BrianSipos: Yes there are definitely cases of recursive use. Tried to minimize this but it just happens when some helper functions do the lock/unlock.

src/refda/adm/ietf_dtnma_agent.h
```diff
@@ -212,6 +212,24 @@ extern "C" {
...
+/// For ./OPER/is-same-ns
+#define REFDA_ADM_IETF_DTNMA_AGENT_ENUM_OBJID_OPER_IS_SAME_NS 40
+/// For ./OPER/match-object-type
+#define REFDA_ADM_IETF_DTNMA_AGENT_ENUM_OBJID_OPER_MATCH_OBJECT_TYPE 42
+/// For ./OPER/match-object-int-range
+#define REFDA_ADM_IETF_DTNMA_AGENT_ENUM_OBJID_OPER_MATCH_OBJECT_INT_RANGE 44
+/// For ./OPER/is-same-object
+#define REFDA_ADM_IETF_DTNMA_AGENT_ENUM_OBJID_OPER_IS_SAME_OBJECT 41
+/// For ./OPER/ref
+#define REFDA_ADM_IETF_DTNMA_AGENT_ENUM_OBJID_OPER_REF 43
```
BrianSipos, line 237: It seems like this `is-same-object` predicate could be rolled into `compare-eq` and `compare-ne` for loose equality of OBJECT to OBJECT. Does that seem to specialized and belongs in this separate oper @justinethier ?

src/refda/eval.c
BrianSipos: This is a good point; is there any downside (other than efficiency) for those cases using `./oper/ref(...)` as the context for label subs?
For the special-purpose filter control parameters, this could be baked into how they operate with "translate" callbacks. For `tbl-filter` this could either be baked in or become advise to the user...

BrianSipos: For general use, the definition of `ref` would need to change to just pass-through literal values. So its signature becomes any -> any.

src/refda/adm/ietf_dtnma_agent.h
BrianSipos: This can be left as a separate predicate for now, writing up the loose compare enhancement separately in https://github.com/JHUAPL-DTNMA/dtnma-adms/issues/68

justinethier:
> There are still FIXME for ACL about modifying the enabled state of rules. This doesn't correlated to any existing permission though.
> 
> @justinethier does it seem like a general permission like `modify-rule` or something more specific like `modify-rule-enabled` is appropriate?

It looks like we went with `modify-rule-enabled` which is fine. Do you have any more questions for me on this one @BrianSipos

BrianSipos:
> > There are still FIXME for ACL about modifying the enabled state of rules. This doesn't correlated to any existing permission though.
> > 
> > @justinethier does it seem like a general permission like `modify-rule` or something more specific like `modify-rule-enabled` is appropriate?
> 
> It looks like we went with `modify-rule-enabled` which is fine. Do you have any more questions for me on this one @BrianSipos 

I don't think so

BrianSipos (commented):
This may deserve a separate PR, but it seems like the ACL tables are sensitive enough to deserve a specific permission such as `modify-acl` with a target referencing one of the EDD tables. The current enforcement is only about `execute` permission on the ACL-associated CTRL objects.

PR #287, opened 2026-05-05
Add sonar job for coverage

BrianSipos:


CMakeLists.txt
```diff
@@ -193,6 +207,8 @@ set(CMAKE_CTEST_ARGUMENTS
   --output-on-failure
 )
 
+# Install config used by tall targets
```
mfarina1, line 210: what's a tall target? might need to elaborate more in this comment

BrianSipos: That's a [CMake term](https://cmake.org/cmake/help/book/mastering-cmake/chapter/Key%20Concepts.html#targets) and this is config for CMake.

mfarina1 (approved):

PR #297, opened 2026-05-06
Refactored mutex lock/unlock behavior for consistency

### LINKED ISSUE #296: Assert on mandatory lock/unlock failures
The current logic of returning an error code is not appropriate for critical errors related to failure of mandatory lock/unlock logic. It will cause false-positive leaks from checkers.

Probably the right thing to do is this, with a consistent failure message:
```c
CACE_LOG_ERR(...)
fprintf(stderr, ...)
static_assert(res == 0)
```

BrianSipos:
This is follow-on to #292.

Closes #296


mfarina1 (commented):
Looks like you just need to fix the formatting (by running the formatting script?)

mfarina1 (approved):

Issue #197, opened 2025-09-05
Improvements to documentation of source code


justinethier:
Feedback from David E.:

- We do a good job of including copyright headers on the top of every file.  This should really be paired with at least a brief description of the file unless obvious from filename or covered by a README

- In dtnma-tools/src for example, it is not clear to me what half of those directories are.  A README with a brief one-line description of each folder would go a long way to helping new developers to the project.

PR #298, opened 2026-05-06
Updating API docs to properly define groups and consistently use config defines

BrianSipos:
This (should be) no change to source code or behavior, just comments and preprocessor mechanism (but not ultimate logic).

Related to, but does not close, issue #197


mfarina1 (approved):
PR is passing SonarQube analysis now, so it looks good to me

Issue #225, opened 2025-12-03
refdm clear_reports endpoint returns error on success


RichLandau1:
The agentEidClearReportsHandler method sends an error response even upon success, it should send an OK with no content.

BrianSipos:
@RichLandau1 can you be more specific about this failure? It appears that a POST request will always succeed (if the agent is valid) but other methods will correctly result in 405 (method not allowed) response. Also, HTTP status 204 is a 2xx series success response for POST so that should be handled as success by a client.

RichLandau1:
@BrianSipos it appears this has been fixed since my report, if you look at the [current iteration](https://github.com/JHUAPL-DTNMA/dtnma-tools/blob/2e04dad2e6dfd68ffc540fa9abe6e4348d0020f8/src/refdm/nm_rest.c#L904) versus [the code at the time of reporting](https://github.com/JHUAPL-DTNMA/dtnma-tools/blob/0276f0ea6ac2830de4171064a239619dd282e586/src/refdm/nm_rest.c) you can see that `mg_send_http_error` was replaced with `mg_response_header_start` and `mg_response_header_send`.

I have not tested it myself since this change but I'll close as OBE.

PR #312, opened 2026-07-27
Cleanup REST status codes

BrianSipos:
This is follow-on to inconsistencies observed when looking into #225.

### DIFF
--- a/src/refdm/nm_rest.c
+++ b/src/refdm/nm_rest.c
@@ -39,8 +39,6 @@
 #define REQUEST_BODY_CHUNK 4096
 
 #define BASE_API_URI      "/nm/api"
-#define VERSION_URI       BASE_API_URI "/version$"
-#define AGENTS_URI        BASE_API_URI "/agents$"
 #define AGENTS_IDX_PREFIX BASE_API_URI "/agents/idx/"
 #define AGENTS_EID_PREFIX BASE_API_URI "/agents/eid/"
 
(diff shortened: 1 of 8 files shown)


mfarina1 (approved):

src/refdm/nm_rest.c
```diff
@@ -247,14 +247,20 @@ static int agentsPostHandler(struct mg_connection *conn)
...
     {
-        m_string_t resp;
-        m_string_init_printf(resp, "Successfully created agent %s", m_string_get_cstr(eid));
-
-        // FIXME should really be a 207
-        mg_send_http_ok(conn, "text/plain", m_string_size(resp));
-        mg_printf(conn, "%s", m_string_get_cstr(resp));
-        m_string_clear(resp);
-        retval = HTTP_OK;
+        mg_response_header_start(conn, HTTP_CREATED);
```
justinethier: POSTs are not necessarily creating new resources in our API. It would probably be better to return 200 (report operations) or 202 (send).

> https://blog.postman.com/what-are-http-status-codes/
> 
> Success codes confirm that the client’s request was received, understood, and successfully processed. These are the responses you want to see when testing your API.
> 
> - 200 OK indicates that the request was successful and the server returned the requested data.
> 
> - 201 Created means the request succeeded and created a new resource. POST requests typically return this code along with a Location header pointing to the newly created resource.
> 
> - 202 Accepted indicates the request has been accepted for processing but hasn’t been completed yet. This status is common for asynchronous operations where immediate processing isn’t possible.
> 
> - 204 No Content confirms that the request was successful, but the server did not return any data. This is typical for successful DELETE operations or PUT requests that don’t need to return data.

justinethier (approved):
Overall looks good, comment could be addressed later.

src/refdm/nm_rest.c
BrianSipos: The registration of new agents here _does_ create new resources, and this response now includes the Location of the agent base URI to construct later `send` and `reports` queries.

BrianSipos: It may actually be a better API to have a POST which _does not_ create a new agent (it's already registered) to simply redirect (303 See Other) to the agent's location. This means that an ignorant API user can simply POST for an agent they want to access, then any valid response (2xx or 3xx) will contain a Location base URI for that agent.

PR #313, opened 2026-07-28
Remove redundant prefix from NM REST API

BrianSipos:
This corresponds to changes for https://github.com/NASA-AMMOS/anms/issues/279


mfarina1 (approved):

justinethier (approved):
Good change. If we are putting together release notes we'll want to make this one prominent since it's a breaking change to the API.

BrianSipos:
> Good change. If we are putting together release notes we'll want to make this one prominent since it's a breaking change to the API.

This is a good point, and can be mitigated with an internal 308 Permanent Redirect, which I will add before merging. But not all clients will follow redirects.

BrianSipos:
To avoid API churn, I am going to try to adjust how ANMS proxies to REFDM rather than change the REFDM paths prefix.

